PR #5, opened 2021-04-12
[CDAP-17731] Add current namespace to footer

elfenheart:
JIRA: https://cdap.atlassian.net/browse/CDAP-17731


app/cdap/components/Footer/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    lineHeight: '53px',
+    margin: '0',
+  },
+}));
+
+const nonNamespacePages = ['Operations', 'Reports', 'Administration'];
+export default function Footer(props) {
+  const footerText = Theme.footerText;
+  const footerUrl = Theme.footerLink;
+  // 'project-id-30-characters-name1/instance-id-30-characters-name';
```
njbriggs, line 75: Nit: Should this stay?

njbriggs (approved):

app/cdap/components/Footer/index.tsx
elfenheart: I think we can keep this, so that we have information about the content if/when we modify this in the future

PR #7, opened 2021-04-15
[CDAP-17870]Partially rewrite ConfigurableTab for ts&css-in-js migration

ajainarayanan:
- Transitions js -> tsx
- Adds types for props and state for ConfigurableTab component

**Note:**
- Not doing transition of styles as the `.cask-confiurable-tab` is being used in way more places. The usage of the css class is also complex enough that it warrants a separate change. 
- Since this causes too much of scope creep I am not doing all those in this PR.
- Not migrating Tab and children components for Tab either. We should ideally use Tab from material-ui and customize it in ways we need.
- We don't achieve consistency with this change but this is one step closer towards that.

### DIFF
--- a/app/cdap/components/ConfigurableTab/index.tsx
+++ b/app/cdap/components/ConfigurableTab/index.tsx
@@ -13,49 +13,61 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Tabs from '../Tabs';
 import TabHeaders from 'components/Tabs/TabHeaders';
 import TabHead from 'components/Tabs/TabHead';
 import TabGroup from 'components/Tabs/TabGroup';
 import classnames from 'classnames';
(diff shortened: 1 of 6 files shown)


app/cdap/components/ConfigurableTab/index.tsx
```diff
@@ -77,10 +89,10 @@ export default class ConfigurableTab extends Component {
       }
       tabs.push(tab);
     });
-    let activeTab = tabs.find((tab) => this.state.activeTab === tab.id);
+    const activeTab = tabs.find((tab) => this.state.activeTab === tab.id);
     return (
-      <div className="cask-configurable-tab">
-        <Tabs layout={this.state.layout}>
+      <div className={`cask-configurable-tab ${this.props.className || ''}`}>
```
elfenheart: use classnames?

elfenheart (approved):
small comment, otherwise 👍  LGTM

ajainarayanan:
Closing and cherry-picking this change to base branch to avoid merge commits

PR #12, opened 2021-04-16
[CDAP-17924] Redesign namespace admin

elfenheart:
JIRA: https://cdap.atlassian.net/browse/CDAP-17924

### DIFF
--- a/app/cdap/components/NamespaceAdmin/ComputeProfiles.tsx
+++ b/app/cdap/components/NamespaceAdmin/ComputeProfiles.tsx
@@ -1,0 +1,75 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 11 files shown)


app/cdap/components/NamespaceAdmin/ComputeProfiles.tsx
```diff
@@ -0,0 +1,75 @@
...
+      display: 'none',
+    },
+    inputLabel: {
+      marginBottom: 0,
+      cursor: 'inherit',
+    },
+  };
+});
+
+const ComputeProfiles: React.FC = () => {
```
ajainarayanan, line 43: nit: If we don't have any props do we need the React.FC type here?

app/cdap/components/NamespaceAdmin/Preferences.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  return (
+    <div>
+      <div className={classes.subtitleSection}>
+        <Button color="primary" variant="contained" onClick={toggleModal}>
+          Edit
+        </Button>
+      </div>
+      <SortableTable
```
ajainarayanan: Can we use the `Table` that we introduced recently for consistency? May be modify `SortableTable` to use the new components, but feel that might be a bigger change.

ajainarayanan (approved):
Minor comments 👍 LGTM

app/cdap/components/NamespaceAdmin/ComputeProfiles.tsx
elfenheart: it's to establish the exported type as a react component instead of normal function

PR #13, opened 2021-04-16
[CDAP-17925] Add drivers list to namespace admin

elfenheart:
JIRA: https://cdap.atlassian.net/browse/CDAP-17925

### DIFF
--- a/app/cdap/components/NamespaceAdmin/store/ActionCreator.ts
+++ b/app/cdap/components/NamespaceAdmin/store/ActionCreator.ts
@@ -18,11 +18,16 @@ import { MyNamespaceApi } from 'api/namespace';
 import { MyPipelineApi } from 'api/pipeline';
 import { MyDatasetApi } from 'api/dataset';
 import { MyPreferenceApi } from 'api/preference';
-import Store, { NamespaceAdminActions } from 'components/NamespaceAdmin/store';
-import { PIPELINE_ARTIFACTS } from 'services/helpers';
+import { MyMetadataApi } from 'api/metadata';
+import { MyArtifactApi } from 'api/artifact';
+import MyCDAPVersionApi from 'api/version';
+import Store, { IDriver, NamespaceAdminActions } from 'components/NamespaceAdmin/store';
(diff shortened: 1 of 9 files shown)


app/cdap/components/NamespaceAdmin/store/ActionCreator.ts
```diff
@@ -106,6 +111,64 @@ export function getPreferences(namespace) {
     });
 }
 
+export function getDrivers(namespace) {
+  const requestNamespace = namespace ? namespace : getCurrentNamespace();
+
+  MyCDAPVersionApi.get()
```
ajainarayanan, line 117: This should be available in the VesionStore? I have made the change in main.js to load rest of the children after the CDAPVersion has been updated in the VersionStore.

app/cdap/components/Popover/index.js
```diff
@@ -81,10 +83,15 @@ export default class Popover extends PureComponent {
   };
 
   componentDidMount() {
+    this._isMounted = true;
```
ajainarayanan: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html We should try to avoid this. Seems like we are missing some to cancel or unsubscribe to some event handler that triggers a setState on an unmounted component.

ajainarayanan (approved):
Minor comments. Otherwise 👍 LGTM

app/cdap/components/NamespaceAdmin/store/ActionCreator.ts
elfenheart: was that a recent change?

ajainarayanan: Sorry I missed that it is available in the feature branch for connection management. https://github.com/cdapio/cdap-ui/pull/16/files#diff-270f5027c1d7025aa0c32f3dd64a385c6143083416f3b3c3bb230874a663416aR206-R214. 

Just noticed that this change is not based off of the feature branch. We can fetch the version from API here for now. Should be fine.

PR #16, opened 2021-04-23
[CDAP-17871] [CDAP-17875] Adds base connections listing + browsing capability with new Connections service

ajainarayanan:
JIRA 
- https://cdap.atlassian.net/browse/CDAP-17871
- https://cdap.atlassian.net/browse/CDAP-17875

Changes
- Adds `GenericBrowser` to handle browsing any connections
- Adds initial POC to use `MemoryRouter` from `react-router-dom` to use same code base for both React and Angular
- Adds explore and list connections API

Auxillary changes
- Changes `main.js` to load CDAPVersion before loading children
- Fixes `Alert` to not setState when unmounted.
- Adds `tsconfig.json` config `downlevelIteration` to support using `Map`s in `for..of` loop 

TODO
- Addresses only happy path for connections browsing.
- Still missing,
  - Breadcrumbs
  - Search/Filter/Sorting in connections browser
  - Actual working connections from pipelines Browse button

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
@@ -1,0 +1,160 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import LoadingSVG from 'components/LoadingSVG';
+import { humanReadableDate, isNilOrEmptyString, objectQuery } from 'services/helpers';
+import { exploreConnection } from 'components/Connections/Browser/GenericBrowser/apiHelpers';
+import * as React from 'react';
+import capitalize from 'lodash/capitalize';
+import If from 'components/If';
+import FolderIcon from '@material-ui/icons/Folder';
+import DescriptionIcon from '@material-ui/icons/Description';
+import makeStyle from '@material-ui/core/styles/makeStyles';
+import Table from 'components/Table';
+import TableHeader from 'components/Table/TableHeader';
+import TableRow from 'components/Table/TableRow';
+import TableCell from 'components/Table/TableCell';
+import TableBody from 'components/Table/TableBody';
+import { getCurrentNamespace } from 'services/NamespaceStore';
+import { useLocation } from 'react-router-dom';
+
+const useStyle = makeStyle(() => {
+  return {
+    nameWrapper: {
+      display: 'flex',
+      gap: '5px',
+    },
+    grid: {
+      height: '100%',
+      marginTop: '15px',
+    },
+    loadingContainer: {
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center',
+      height: '100%',
+    },
+  };
+});
+
+const ICON_MAP = {
+  directory: <FolderIcon />,
+  file: <DescriptionIcon />,
+};
+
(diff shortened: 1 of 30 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+    } catch (e) {
+      setLoading(false);
+      setError(`Failed to explore connection : "${e.message}"`);
+    }
+  };
+  const onExplore = (entityName) => {
+    if (path === '/') {
+      setPath(`/${entityName}`);
+    } else {
+      setPath(`${path}/${entityName}`);
```
njbriggs, line 85: Should this use `path` on the selected entity rather than constructing a path on the client?

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+              key={i}
+              to={`/ns/${getCurrentNamespace()}/connections/${selectedConnection}?path=${getPath(
+                entity.name
+              )}`}
+              onClick={() => onExplore(entity.name)}
+            >
+              <TableCell>
+                <div className={classes.nameWrapper}>
+                  <If condition={entity.type === 'directory'}>
+                    <FolderIcon />
```
njbriggs: Nit: we need to a way to determine if other icons should be shown, like a database icon.

app/cdap/components/Connections/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <If condition={!enableRouting}>
+        <MemoryRouter initialEntries={[`/ns/${getCurrentNamespace()}/connections`]}>
+          <ConnectionRoutes enableRouting={enableRouting} />
+        </MemoryRouter>
+      </If>
+    </React.Fragment>
+  );
+}
+
+(Connections as any).propTypes = {
```
njbriggs: Why do we need to set `propTypes` like this?

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
ajainarayanan: The selected entity has the name. If this is one of the root folders it will have a preceeding forward slash otherwise the new path will be existing path + the entity name.

Eg:
```
- dir1
- dir2
- dr3
```
clicking on `dir1` will generate a path `/dir1`

If under `dir1` the contents are

```
- dir12
- dir34
- dir56
```
and clicking on `dir34` will generate `/dir1/dir34`. Here `dir34` is the entity name.

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
ajainarayanan: You are correct. This has to be dynamic for each connection. Right now I am not addressing the icons. Just the base structure. We need to think about icons for categories, connectors, connections and individual entities in each connection. 

The good part is the API already has the information we need for showing icons under a connection while exploring. We will need to address this when we deal with icons as a whole for connections/wrangler interaction.

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+      const res = await exploreConnection({
+        connectionid: selectedConnection,
+        path,
+      });
+      setEntities(res.entities);
+      setLoading(false);
+    } catch (e) {
+      setLoading(false);
+      setError(`Failed to explore connection : "${e.message}"`);
+    }
```
antony-britto, line 79: finally {
  setLoading(false);
}

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  if (error) {
+    throw error;
+  }
+  if (!Array.isArray(entities) || (Array.isArray(entities) && !entities.length)) {
+    return <div>No entities available</div>;
+  }
+  let headers = ['Name', 'Type'];
+  const properties = objectQuery(entities, 0, 'properties');
+  headers = [...headers, ...properties.map((prop) => capitalize(prop.key))];
```
elfenheart, line 119: is it possible that there is no `entities` (ie. empty folder), the `...properties.map` will fail in that case

PR #21, opened 2021-05-14
[CDAP-17988] Breadcrumb for new generic browser

njbriggs:
Jira: https://cdap.atlassian.net/browse/CDAP-17988

This uses material-ui's Breadcrumbs to add a breadcrumb for the generic browser. It uses react-router to change locations and has the look and feel of the current breadcrumb, with the exception of the collapsed state.

Base state:
image

A couple levels in:
image

Collapsed with > 3 levels:
image

Note the different style. I don't know if the appearance can be overridden. Also, clicking it will expand the directories in-place rather than showing a menu. If this is too big a change, we'll need to create our own bread crumb.

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/Breadcrumb.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/Breadcrumb.tsx
@@ -1,0 +1,66 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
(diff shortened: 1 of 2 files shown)


elfenheart (approved):

njbriggs:
Cherry-picked into https://github.com/cdapio/cdap-ui/tree/bugfix-ui/CDAP-17870 after approval.

PR #27, opened 2021-05-20
Fix E2E tests and disable current invalid tests

elfenheart:
Disabling wrangler tests since those are not currently valid with the new connection manager.
Disabling hierarchy widget, currently being fixed
Fixed secure keys test and Alert component


njbriggs (approved):
Can you file a bug to re-enable the Wrangler tests?

elfenheart:
https://cdap.atlassian.net/browse/CDAP-18005

PR #34, opened 2021-05-27
Connection management modify table to use common table component

elfenheart:
image


lahwran (approved):
Looks generally reasonable to me. I can't verify the details of the usage of the table component. If you feel it's low risk, I'd say merge.

PR #36, opened 2021-06-01
[CDAP-18038] Empty and error messaging to match app standard

njbriggs:
Switched to use EmptyMessageContainer and ErrorBanner

Jira: https://cdap.atlassian.net/browse/CDAP-18038

Empty state:
image

Error state:
image

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
@@ -15,17 +15,21 @@
  */
 
 import * as React from 'react';
-import MuiAlert from '@material-ui/lab/Alert';
 import { isNilOrEmptyString } from 'services/helpers';
 import { exploreConnection } from 'components/Connections/Browser/GenericBrowser/apiHelpers';
 import debounce from 'lodash/debounce';
 import makeStyle from '@material-ui/core/styles/makeStyles';
+import T from 'i18n-react';
 import { getCurrentNamespace } from 'services/NamespaceStore';
 import { useLocation } from 'react-router-dom';
 import Breadcrumb from 'components/Connections/Browser/GenericBrowser/Breadcrumb';
 import SearchField from 'components/Connections/Browser/GenericBrowser/SearchField';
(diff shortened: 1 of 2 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -141,15 +145,23 @@ export function GenericBrowser({ selectedConnection }) {
...
-      <If condition={isEmpty && !error}>
-        <MuiAlert severity="info" className={classes.alert}>
-          No entities available
-        </MuiAlert>
+      <If condition={isEmpty && !loading}>
+        <EmptyMessageContainer title="No entities available">
+          <ul>
+            <li>
+              <span className="link-text" onClick={clearSearchString}>
+                {T.translate(`features.EmptyMessageContainer.clearLabel`)}
```
lahwran, line 152: What makes the difference between this string, which gets translated, and the strings on lines 155 and 158, which currently do not? Is this intentional?

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -141,15 +145,23 @@ export function GenericBrowser({ selectedConnection }) {
           loading={loading}
         />
       </If>
-      <If condition={isEmpty && !error}>
-        <MuiAlert severity="info" className={classes.alert}>
-          No entities available
-        </MuiAlert>
+      <If condition={isEmpty && !loading}>
```
lahwran, line 147: Should this be visible in the event of an error which is shown in the ErrorBanner component below, on line 164?

njbriggs: Yes, this matches the current behavior.

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
njbriggs: This should all be internationalized. Good catch.

njbriggs: Updated with full I18n for this portion.

ajainarayanan (approved):

PR #37, opened 2021-06-03
[CDAP-17867] Pushdown Config UI

lahwran:
https://cdap.atlassian.net/browse/CDAP-17867

Add ConfigurationGroup-based configuration ui tab to pipeline config modal in studio.

Screen Shot 2021-06-03 at 1 14 46 PM


nitinmotgi:
@bdmogal What is the language here ? Why do we need so many configuration for ELT push-down.

app/cdap/components/PushdownConfig/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    order: 'DESC',
+  };
+
+  return MyPipelineApi.getPluginProperties(pluginParams).map(([res]) => {
+    return res;
+  });
+}
+interface IPushdownConfig {
+  enabled: boolean;
+  transformationPushdown: null | {
```
elfenheart: nullable on interface can be a `?`

```
transformationPushdown?: {
  plugin?: object
}
```

app/cdap/components/PushdownConfig/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    plugin: null | object;
```
elfenheart: ideally this should be more descriptive than `object`

app/cdap/components/PushdownConfig/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const plugin = objectQuery(transformationPushdown, 'plugin');
+  const valueProperties = objectQuery(plugin, 'properties') || {};
+
+  const classes = useStyles();
+  const onChange = (newProps) => {
+    const newTransformationPushdown = {
+      plugin: {
+        ...plugin,
+        properties: {
+          ...valueProperties,
```
elfenheart: this should just be `...newProps`, the onChange argument will get the entire value

app/directives/my-pipeline-configurations/my-batch-pipeline-config/my-batch-pipeline-config.html
```diff
@@ -201,6 +206,13 @@
           store="BatchPipelineConfigCtrl.store">
         </my-post-run-actions>
       </div>
+      <div class="configuration-step-content" id="pushdown-tab-content"
+        ng-if="BatchPipelineConfigCtrl.activeTab === 'pushdown'" style="height: 100%; width: 100%; display: block;">
+        <pushdown-config value="BatchPipelineConfigCtrl.pushdownConfig"
+          on-value-change="BatchPipelineConfigCtrl.onPushdownConfigChange"
+          style="height: 100%; width: 100%; display: block;">
```
elfenheart: don't do inline styles, you should be able to apply the styles with css

app/hydrator/services/create/stores/config-store.js
```diff
@@ -111,6 +111,13 @@ class HydratorPlusPlusConfigStore {
         this.setNumRecordsPreview(this.state.config.numOfRecordsPreview);
         this.setMaxConcurrentRuns(this.state.config.maxConcurrentRuns);
       }
+      this.setPushdownConfig({
+        // null: getDefaultConfig, undefined: imported json, {}: just in case
+        enabled: (this.state.config.transformationPushdown !== null
+          && this.state.config.transformationPushdown !== undefined
```
elfenheart: i think in this case, doing `!this.state.config.transformationPushdown` would suffice.

btw, why would imported JSON have `{}`?

lahwran: Hmm, I don't remember why I was thinking that. I think it may have been caused by something I was doing while testing, before I'd figured out the dataflow into config-store.js. Simplification reasonable, will remove.

app/cdap/components/PushdownConfig/index.tsx
lahwran: Ah, good to know

app/hydrator/services/create/stores/config-store.js
lahwran: Looks like in javascript, `(!null) === true`, so I can only remove the object check.

lahwran: I just realized I'm an idiot. You meant `!!`, didn't you? Of course `(!null) === true`.

elfenheart: sorry yeah, i meant `!!`

app/directives/my-pipeline-configurations/my-batch-pipeline-config/my-batch-pipeline-config.html
```diff
@@ -206,11 +206,11 @@
...
         </my-post-run-actions>
       </div>
-      <div class="configuration-step-content" id="pushdown-tab-content"
-        ng-if="BatchPipelineConfigCtrl.activeTab === 'pushdown'" style="height: 100%; width: 100%; display: block;">
+      <div class="configuration-step-content pushdown-config-full-block" id="pushdown-tab-content"
+        ng-if="BatchPipelineConfigCtrl.activeTab === 'pushdown'">
         <pushdown-config value="BatchPipelineConfigCtrl.pushdownConfig"
           on-value-change="BatchPipelineConfigCtrl.onPushdownConfigChange"
-          style="height: 100%; width: 100%; display: block;">
+          class="pushdown-config-full-block">
```
elfenheart, line 213: just verifiying, is this class required at both levels?

lahwran: It is. It's to inflate/limit them to the full width and height of the parent, so that the styling in the react component can respond to that height to do scrolling without simply overflowing the configuration popup. An alternative could be to expand the entire configuration popup dynamically - it didn't happen naturally, and I didn't try to scope it, might not be too bad. This way was an easier path and I was being a greedy planning algorithm at the time.

elfenheart (approved):

elfenheart:
👍  LGTM, please squash the commits

PR #38, opened 2021-06-04
[CDAP-17874] Add ability to edit, clone, export, and delete connectio…

elfenheart:
…n from connection list view

### DIFF
--- a/app/cdap/components/Connections/Browser/SidePanel/CategorizedConnections.tsx
+++ b/app/cdap/components/Connections/Browser/SidePanel/CategorizedConnections.tsx
@@ -24,12 +24,20 @@ import { Link } from 'react-router-dom';
 import { getCurrentNamespace } from 'services/NamespaceStore';
 import If from 'components/If';
 import classnames from 'classnames';
+import ActionsPopover, { IAction } from 'components/ActionsPopover';
+import { objectQuery } from 'services/helpers';
+import DownloadFile from 'services/download-file';
+import ConfirmationModal from 'components/ConfirmationModal';
+import CreateConnectionModal from 'components/Connections/CreateConnectionModal';
+import { ConnectionsApi } from 'api/connections';
 
 interface ICategorizedConnectionsProps {
   categorizedConnections: Map<string, any[]>;
   categories: string[];
   onConnectionSelection: (conn: string) => void;
   selectedConnection: string;
+  boundaryElement: any;
+  fetchConnections: () => void;
 }
 const Accordion = withStyles({
   root: {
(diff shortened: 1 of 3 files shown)


app/cdap/components/Connections/Browser/SidePanel/CategorizedConnections.tsx
```diff
@@ -156,6 +183,105 @@ export function CategorizedConnections({
     }
   };
 
+  function getConnectionConfig(conn) {
```
lahwran: Nit: I would personally prefer for functions that don't depend on hooks to be at the top level, rather than nested.

app/cdap/components/Connections/Browser/SidePanel/CategorizedConnections.tsx
```diff
@@ -156,6 +183,105 @@ export function CategorizedConnections({
...
+      category: conn.plugin.category,
+      plugin: conn.plugin,
+    };
+
+    return connectionConfig;
+  }
+
+  function exportConnection(conn) {
+    const connectionConfig = getConnectionConfig(conn);
+    const artifact = objectQuery(conn, 'plugin', 'artifact') || {};
```
lahwran: Comment only: `conn?.plugin?.artifact || {}` will preserve types more accurately. Might be possible to get objectQuery to preserve the types, not sure how much typescript's type system allows there.

app/cdap/components/Connections/Browser/SidePanel/CategorizedConnections.tsx
```diff
@@ -156,6 +183,105 @@ export function CategorizedConnections({
...
+  }
+
+  function cloneConnection(conn) {
+    const config = getConnectionConfig(conn);
+    delete config.name;
+
+    openConnectionConfig(config);
+  }
+
+  function openConnectionConfig(config) {
```
lahwran, line 215: Nit: With this many handlers, I would prefer to move the state management into a useReducer, to isolate local state updates from actions taken in response to state. I don't it's worth holding up to refactor, perhaps more of a comment for next large component.

app/cdap/components/Connections/Browser/SidePanel/CategorizedConnections.tsx
```diff
@@ -156,6 +183,105 @@ export function CategorizedConnections({
...
+    hide: {
+      enabled: false,
+    },
+  };
+
+  let confirmDeleteElem;
+  if (connectionToDelete) {
+    confirmDeleteElem = (
+      <div>
+        Are you sure you want to delete{' '}
```
lahwran, line 277: Is translation needed?

elfenheart: we haven't migrate connection management to use i18n yet

lahwran (approved):

PR #39, opened 2021-06-04
[CDAP-17878] Integrates Wrangler UI with V2 wrangler API

ajainarayanan:
JIRA - https://cdap.atlassian.net/browse/CDAP-17878

**Functional change**
- From a user perspective nothing changes. They will still see valid connections/wrangler workspace/ apply directives.
- Behind the scenes
  - The PR primarily addresses the integration of wrangler UI with v2 wrangler API.
  - This should enable a working solution of wrangler from wrangler/connections side. User should be able to navigate from connection to wrangler and apply directives.
  - This also fixes the "Wrangle" button in the wrangler transform which does the same thing.
  - Fixes DataprepActionCreator based on changes in response from v2 API

**Code Improvements**
- Cleans up `Connections`. Removes `enableRouting` and `singleWorkspaceMode` and consolidates them to `mode`.
- Introduces Context to be shared between Connection and children of Connection component.

**TODO**
- The 'Browse' functionality is still not addressed in this PR. The specification API from backend is available and we should be able to add it as a separate PR
- `DataPrepHome`, `DataPrep` modules can be converted to tsx. I have not done that in this PR as that conversion introduces like 100+ errors which increases the scope of this PR. Once we do this we should be able to remove the hardcoding `ROUTED_WORKSPACE` and `INMEMORY`

### DIFF
--- a/app/cdap/components/DataPrep/TopPanel/index.js
+++ b/app/cdap/components/DataPrep/TopPanel/index.js
@@ -20,8 +20,6 @@ import Loadable from 'react-loadable';
 import LoadingSVGCentered from 'components/LoadingSVGCentered';
 import DataPrepStore from 'components/DataPrep/store';
 import { objectQuery, preventPropagation } from 'services/helpers';
-import cdapavsc from 'services/cdapavscwrapper';
-import { directiveRequestBodyCreator } from 'components/DataPrep/helper';
 import NamespaceStore from 'services/NamespaceStore';
 import MyDataPrepApi from 'api/dataprep';
 import T from 'i18n-react';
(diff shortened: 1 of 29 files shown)


app/cdap/components/DataPrep/TopPanel/index.js
```diff
@@ -382,14 +366,14 @@ export default class DataPrepTopPanel extends Component {
             <span className="text-danger">{this.state.onSubmitError}</span>
           ) : null}
           {this.state.higherVersion ? this.renderUpgradeBtn() : null}
-          {this.props.singleWorkspaceMode ? this.renderApplyBtn() : null}
-          {!this.props.singleWorkspaceMode ? (
+          {this.props.mode === 'ROUTED_WORKSPACE' ? this.renderApplyBtn() : null}
```
njbriggs, line 369: For all uses of these constants: Should this be imported from the enum?

ajainarayanan: @njbriggs I added a note in the PR description for this. Ideally we should. But for us to be able to use enum we need to change this to a tsx file. If I do that there are 100+ errors which is quite a bit of change for this PR. 

This should be done but outside of this PR.

njbriggs (approved):

PR #40, opened 2021-06-07
 [CDAP-18040] Type-specific formatting for connection browser table

njbriggs:
This adds DataFormatter, which I'm proposing to use for all data fields that are returned from the back end. We should also review the formats used to display values.

Jira: https://cdap.atlassian.net/browse/CDAP-18040


njbriggs:
Replaced by #50

PR #44, opened 2021-06-16
[CDAP-17878] Add ability to browse connection from pipeline

elfenheart:
JIRA: https://cdap.atlassian.net/browse/CDAP-17878

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
@@ -19,6 +19,7 @@ import { isNilOrEmptyString } from 'services/helpers';
 import {
   exploreConnection,
   createWorkspace,
+  getPluginSpec,
 } from 'components/Connections/Browser/GenericBrowser/apiHelpers';
 import capitalize from 'lodash/capitalize';
 import countBy from 'lodash/countBy';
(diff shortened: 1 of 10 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -116,7 +119,11 @@ export function GenericBrowser({ selectedConnection }) {
     const { name: entityName, canBrowse } = entity;
     if (!canBrowse) {
       setLoading(true);
-      return onCreateWorkspace(entity);
+      getEntitySpec(entity);
```
lahwran: Question: Does this function have side effects? If so, how would you feel about renaming it away from `get...`, eg perhaps `retrieveEntitySpec` or `loadEntitySpec`? should it be duplicated on this line vs the one below?

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -116,7 +119,11 @@ export function GenericBrowser({ selectedConnection }) {
...
+
+      return typeof onEntitySelect === 'function'
+        ? getEntitySpec(entity)
+        : onCreateWorkspace(entity);
```
lahwran: onCreateWorkspace isn't a function name I'd expect to return a value. What kind of value does it return? If it's a promise with no return value, I think it would be clearer to do:
```tsx
const onExplore = async (entity) => {
  ...;
  if (typeof onEntitySelect === 'function') {
    await getEntitySpec(entity);
  } else {
    await onCreateWorkspace(entity);
  }
}
```

app/cdap/components/DataPrepConnections/PluginConnectionBrowser/index.tsx
```diff
@@ -76,250 +69,39 @@ class PluginConnectionBrowser extends React.PureComponent<
     showBrowserModal: false,
     error: null,
     loading: false,
-    defaultConnectionId: null,
-    defaultConnectionType: null,
+    connectionName: null,
   };
 
-  private ee = ee(ee);
+  public eventEmitter = ee(ee);
```
lahwran, line 75: eeeeeeeeeeeeeeeee

lahwran (approved):

PR #49, opened 2021-06-17
[CDAP-17510] Disables checkbox for selecting column on table replication if primary key

seanfeld:
Bugfix

Fixes [CDAP-17510](https://cdap.atlassian.net/browse/CDAP-17510).

Disables checkbox for selecting column on table replication if primary key. 

I have a couple of questions though:
- Should we only disable it if it is selected? Is it possible for a user to deselect it in another way?
- Should we notify the user why this checkbox is disabled? We can use tooltips as long as the tooltip is accessible?

Screen Shot 2021-06-16 at 9 27 25 PM

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SelectColumns/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectColumns/index.tsx
@@ -15,6 +15,7 @@
  */
 
 import * as React from 'react';
+import T from 'i18n-react';
 import withStyles, { WithStyles, StyleRules } from '@material-ui/core/styles/withStyles';
 import { createContextConnect } from 'components/Replicator/Create';
 import { Map } from 'immutable';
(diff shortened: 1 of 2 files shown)


app/cdap/components/Replicator/Create/Content/SelectColumns/index.tsx
```diff
@@ -384,6 +384,7 @@ class SelectColumnsView extends React.PureComponent<ISelectColumnsProps, ISelect
                       <Checkbox
                         color="primary"
                         className={classes.radio}
+                        disabled={this.state.primaryKeys.indexOf(row.name) !== -1}
```
elfenheart: Can you check if you click on the select all at the table header, will it deselect the table? I think it will.

elfenheart:
`Should we notify the user why this checkbox is disabled? We can use tooltips as long as the tooltip is accessible?`
That's a good idea, you can add a title prop on it explaning why it's disabled

elfenheart (approved):
👍  LGTM. Please squash and merge

PR #52, opened 2021-06-17
[CDAP-18069] Test connection during connection creation with new API

njbriggs:
Jira: https://cdap.atlassian.net/browse/CDAP-18069

Test in progress:
image

Test success:
image

Test failed:
image

Server-side error during test:
image

### DIFF
--- a/app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
+++ b/app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
@@ -19,13 +19,26 @@ import ConfigurationGroup from 'components/ConfigurationGroup';
 import { Button } from '@material-ui/core';
 import makeStyle from '@material-ui/core/styles/makeStyles';
 import PropertyRow from 'components/ConfigurationGroup/PropertyRow';
+import If from 'components/If';
+import LoadingSVG from 'components/LoadingSVG';
 
-const useStyle = makeStyle(() => {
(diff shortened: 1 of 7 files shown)


app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
```diff
@@ -27,17 +29,30 @@ const useStyle = makeStyle(() => {
...
+    connectionTestMessage: {
+      display: 'flex',
+      height: '2rem',
+      padding: '0 10px',
+    },
+    connectionTestSuccess: {
+      color: '#4ab63c',
+    },
+    connectionTestFailure: {
+      color: '#d15668',
```
elfenheart: are these not theme colors?

elfenheart (commented):
also curious, is there no field level errors like in pipelines?

app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
njbriggs: Looks like the failure color is `red-03`. The green color is not in the theme but it looks like they're not in sorted order so I'll add it.

njbriggs:
Using colors from theme now.

elfenheart (approved):
Just question around whether there is field level error or not? From the screenshots it seems like it's only a global error

PR #53, opened 2021-06-18
[CDAP-17867] Add pushdown ui to post-deploy ui, remove from studio

lahwran:
https://cdap.atlassian.net/browse/CDAP-17867

- Remove from studio
- Add to post-deploy ui
- Use current cloud artifact info from pipeline if present

Screen Shot 2021-06-18 at 1 00 12 AM


app/cdap/components/PushdownConfig/index.tsx
```diff
@@ -105,17 +105,19 @@ interface IPushdownConfig {
...
   const [loading, setLoading] = useState(true);
   const [pluginInfo, setPluginInfo] = useState(null);
   const [pluginWidget, setPluginWidget] = useState(null);
   const pluginProperties = objectQuery(pluginInfo, 'properties') || {};
-  const { enabled, transformationPushdown } = value;
+  const { pushdownEnabled, transformationPushdown } = value;
 
   const plugin = transformationPushdown?.plugin;
   const valueProperties = plugin?.properties || {};
+  console.log("PushdownConfig render", { value, cloudArtifact });
```
seanfeld: do we want to leave this console log?

lahwran: Whoops, removed.

app/cdap/components/PipelineConfigurations/ConfigurationsContent/PushdownTabContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+      });
+    },
+    [dispatch]
+  );
+  return (
+    <div className="configuration-step-content configuration-content-container">
+      <PushdownConfig value={value} onValueChange={onChange} cloudArtifact={cloudArtifact} />
+    </div>
+  );
+}
```
elfenheart: nit: newline at EOF. I think there's a setting in VScode that will add this automatically when you save

app/cdap/components/PushdownConfig/index.tsx
```diff
@@ -167,7 +168,7 @@ export default function PushdownConfig({ value, onValueChange }: IPushdownProps)
       }
     );
     // pass etc
-  }, []);
+  }, [cloudArtifact?.name, cloudArtifact?.scope, cloudArtifact?.version]);
```
elfenheart: I think in this case you can just use `cloudArtifact`? When there's a change in one of these values, a new object is created right?

app/cdap/components/PipelineConfigurations/TabConfig.js
```diff
@@ -60,6 +61,13 @@ const TabConfig = {
       contentClassName: 'pipeline-configurations-body',
       paneClassName: 'configuration-content',
     },
+    {
```
elfenheart: Might have to add a conditional here. We don't want to show this setting on realtime pipeline

app/cdap/components/PushdownConfig/index.tsx
lahwran: Actually yeah, you're right - I was guarding against risk of the object being recreated in PushdownTabContent, but on re-check, it's not. I'll change that

app/cdap/components/PipelineConfigurations/index.js
```diff
@@ -126,28 +177,34 @@ export default class PipelineConfigurations extends Component {
...
-      tabConfig.tabs = TabConfig.tabs.filter((tab) => {
-        return tab.name !== alertsTabName;
-      });
+      tabs = [
+        TABS.computeConfig,
+        TABS.pipelineConfig,
+        TABS.engineConfig,
+        TABS.resources,
+        // no alerts tab for realtime pipelines
+        TABS.pushdown,
```
elfenheart: shouldn't have pushdown config either

elfenheart (approved):

PR #59, opened 2021-06-22
[CDAP-17679] Pluralize overview page replications

lahwran:
https://cdap.atlassian.net/browse/CDAP-17679


njbriggs:
You can see an example of pluralization with i18n here: https://www.npmjs.com/package/i18n-react

You can look in `text-en.yml` for some usage within CDAP.

PR #63, opened 2021-06-24
Bugfix UI/cdap 17806 - Save more often in the replication create flow.

seanfeld:
### Pr Type
**Bugfix**

### Context
[asana task](https://cdap.atlassian.net/browse/CDAP-17806)
TLDR: this pr adds saving to the previous button as well as a useEffect cleanup and component will unmount. 

### Logic
useRef/useEffect cleanup to mimic component will unmount - useEffect cleanups with an empty dependency array doesn't have access to the updated state so you have to use refs to grab the updated values.

Outdated:
So I was unsure how to approach this one. Basically this the asana task said that we needed to save if the user navigates away while in the create flow. It also meant that we needed to save if the user clicked previous, per a conversation with Bhooshan.  So basically I needed to save more often to the state in order to save the users work.

The two best options I thought of in this situation would be to either continue along with the previous logic, as well as adding a save cleanup on a use effect or component will unmount OR saving more often to the parent state with a debounce (this is what I chose in this PR). Adding a save on the navigate + previous clicks was messy because I needed to pass around save functions depending on the step that needed child component state as well.

Unsure about side effects and edge cases but I've done a decent amount of manual testing and it seems to work fine.


app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
```diff
@@ -91,11 +91,12 @@ const PluginConfigView: React.FC<IPluginConfigProps> = ({
   parentArtifact,
   pluginType,
 }) => {
-  const [view, setView] = React.useState(VIEW.configuration);
-  const [values, setValues] = React.useState(pluginConfig || {});
-  const [loading, setLoading] = React.useState(false);
+  const [view, setView] = useState(VIEW.configuration);
+  const [values, setValues] = useState(pluginConfig || {});
+  const [loading, setLoading] = useState(false);
+  const debounce = useDebounce(100);
```
lahwran: this feels like an unnatural api to me because this debounce function does not integrate its debouncing effect into the react render process, the way eg useCallback or useEffect does. I might make a `useDebouncedEffect(delay, effect, keys)` or `const debouncedFunc = useDebounce(delay, func);`. The benefit is that the binding of the function that will be debounced is moved inside the debounce hook, removing any way to express calling the same `debounce(func)` with two different functions in different places.

That said, useDebounce tends to be one of my favorite hooks to have available, so I'm glad to see an implementation of it is available!

app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
```diff
@@ -14,11 +14,11 @@
  * the License.
  */
 
-import * as React from 'react';
+import React, { useEffect, useState } from 'react';
```
lahwran: +1 using direct imports of hooks

app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
```diff
@@ -236,7 +242,7 @@ const PluginConfigView: React.FC<IPluginConfigProps> = ({
         <LoadingSVG />
       </If>
 
-      <StepButtons onNext={handleNext} nextDisabled={isNextDisabled()} />
+      <StepButtons nextDisabled={isNextDisabled()} />
```
lahwran: Nit: Removing onNext seems like it creates a (potentially too mild to care) timing error, depending on whether the debounce function is guaranteed to run after a component gets unmounted. There's always the network to create timing errors, so I'll leave it up to you whether this is an issue at all.

app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
```diff
@@ -169,7 +176,7 @@ const PluginConfigView: React.FC<IPluginConfigProps> = ({
     return isPropertyFilled.filter((propertyValue) => !propertyValue).length > 0;
   }
 
-  function handleNext() {
+  function returnFilteredValues() {
```
lahwran: nit 1: This looks like a useEffect dependencies lint error to me. I'll leave this up to you consider moving this into a `const filterValues = useCallback(values => ..., [pluginWidget])`, then I'd `useEffect(() => ..., [filterValues, values])`, to ensure that the filterValues function is integrated into effect causality. Or even just `const filteredValues = useMemo(() => ..., [pluginWidget, values])`.

nit 2: I would personally name this function as a verb on what it does to the data, since it's doing a pure calculation. To my mental natural language parser, "return" in the function name implies that the function will cause something else outside itself to return, not merely that the rest of the name is what it's returning.

lahwran (approved):
No critical issues, but several suggestions. Change if my arguments convinced you.

lahwran (changes requested):
Ah, it seems I shouldn't have made that an approval yet.

app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
seanfeld: I think my intention was for people to wrap it in a useEffect afterwards and take advantage of the dependency management; so the debounce was only meant to be a utility function that implemented debounce. 

```
useEffect(() => {
  debouncedFn(fn(pieceOfState));
}, [pieceOfState]);
```

I think the problem was I should've named it something like `handleDebounce` instead of `useDebounce` since I wanted to wrap it around a useEffect afterwards.

That being said, i've removed it because i'm not using it anymore but in the future we can make it something more agreeable :) . Maybe something like the `useDebouncedEffect` that you mentioned might make the most sense.

app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
```diff
@@ -182,11 +196,10 @@ const PluginConfigView: React.FC<IPluginConfigProps> = ({
       processedConfigurationGroup.configurationGroups,
       pluginWidget,
       pluginProperties,
```
seanfeld, line 198: @elfenheart is it possible for plugin properties to change here? its based on plugin info, which is a prop. Unsure if i need to make a ref for plugin info as well?

app/cdap/components/Replicator/Create/Content/PluginConfig/index.tsx
seanfeld: hmm nit1 is a bit confusing - renamed to filtered values though

app/cdap/components/Replicator/Create/Content/Advanced/index.tsx
```diff
@@ -141,15 +141,26 @@ const AdvancedView: React.FC<ICreateContext & WithStyles<typeof styles>> = ({
   numInstances,
   setAdvanced,
 }) => {
-  const [localNumInstances, setLocalNumInstances] = React.useState(numInstances || 1);
-  const [taskSelection, setTaskSelection] = React.useState(TASK_OPTIONS.calculate);
-  const [dataAmount, setDataAmount] = React.useState(1);
+  const [localNumInstances, setLocalNumInstances] = useState(numInstances || 1);
+  const [taskSelection, setTaskSelection] = useState(TASK_OPTIONS.calculate);
+  const [dataAmount, setDataAmount] = useState(1);
+  const saveRef = useRef(localNumInstances);
```
elfenheart, line 147: why is useRef necessary here? isn't useRef usually used to reference an element?

app/cdap/components/Replicator/utilities/index.ts
```diff
@@ -383,3 +384,23 @@ export function getTableDisplayName(row) {
...
+ *  const debounce = useDebounce(100);
+ *
+ *  debounce(() => {
+ *   functionToExecute();
+ *  });
+ * @param ms time in ms
+ * @returns function that takes a function as its argument and executes after the debouncea
+ * eg
+ */
+export const useDebounce = (ms: number) =>
```
elfenheart: I think this is not needed anymore?

seanfeld: oh weird i thought i removed it - my bad

app/cdap/components/Replicator/Create/Content/Advanced/index.tsx
seanfeld: useEffect cleanup functions don't have access to the most recent state

app/cdap/components/Replicator/utilities/index.ts
seanfeld: to clarify - with an empty dependency array and mimicking componentWillUnmount it doesn't have access to the newest state

app/cdap/components/Replicator/Create/Content/Advanced/index.tsx
seanfeld: to clarify - with an empty dependency array and mimicking componentWillUnmount it doesn't have access to the newest state

elfenheart: I feel like useRef is generally used for DOM element reference. Perhaps what you want is useCallback instead?

seanfeld: @elfenheart in order to get access to the most recent state in the "unmount" callback you need to make a reference to that value in the state. Its similar to this pr that @lahwran made here https://github.com/cdapio/cdap-ui/pull/92/files - its a pretty widely used pattern with examples in other places online (that I can find if necessary) - I can update this code to use @lahwran 's unmount callback if needed but that code was written after this pr.

elfenheart (approved):

lahwran (approved):
Undoes my "changes requested" to match @elfenheart's approval

PR #79, opened 2021-06-30
CDAP-17946 / check if errors are arrays before forEaching them

seanfeld:
### Bugfix
[jira](https://cdap.atlassian.net/browse/CDAP-17946)

Was unable to reproduce but the error is caused by networkErrors not being an array. Earlier in the code it attempts to define networkErrors `err || []` but that only works if `err` evaluates to be falsey. Should we add logging or something to figure out what networkErrors could be returned as?

### DIFF
--- a/app/cdap/services/helpers.js
+++ b/app/cdap/services/helpers.js
@@ -526,24 +526,38 @@ function categorizeGraphQlErrors(error) {
   if (graphQLErrors.length === 0 && networkErrors.length === 0 && error) {
     errorsByOrigin[GENERIC_ERROR_ORIGIN] = error.message;
   }
+  /* 
+    TODO: These array checks are guarding against the page crashing when graphQL errors
+    or network errors aren't an array - we haven't been able to repro this locally so
+    we don't know how to handle it properly. When are these not an array?
+  */ 
+  if (Array.isArray(graphQLErrors)) {
+    graphQLErrors.forEach(error => {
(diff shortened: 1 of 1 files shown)


app/cdap/services/helpers.js
```diff
@@ -527,23 +527,29 @@ function categorizeGraphQlErrors(error) {
...
-  graphQLErrors.forEach(error => {
-    const errorOrigin = objectQuery(error, 'extensions', 'exception', 'errorOrigin') || GENERIC_ERROR_ORIGIN;
-    if (errorsByOrigin.hasOwnProperty(errorOrigin)) {
-      errorsByOrigin[errorOrigin].push(error.message);
-    }
-    else {
-      errorsByOrigin[errorOrigin] = [error.message];
-    }
-  });
+  if (Array.isArray(graphQLErrors)) {
```
lahwran, line 534: Since this is a fix for an error we're not sure how to reproduce, I'd suggest adding a comment saying so here. `// TODO: Array.isArray() is guarding against a bug we haven't reproed. error is the {error} value from apollo.useQuery(). When is this false?` - I also might add an `else { console.error(...); }`. I don't think it's the worst thing to commit a partial fix like this, but I think it's important that the code not be silent about the reasoning behind it.

lahwran (changes requested):
I'd feel better approving this if the code contains a comment explaining our current state of knowledge.

seanfeld:
@lahwran can you relook at this?

lahwran (approved):
Whoops, sorry to leave you hanging for a couple of days!

PR #92, opened 2021-07-07
[CDAP-18173] Fix unmount bug in ConfigurationGroup

lahwran:
This was causing the PushdownConfig settings to get partially reset when closing the settings modal.

Questions for reviewer:
- I considered refactoring ConfigurationGroup to not use its existing ref workaround, but decided that the possibility of unexpected dependence on the current ref behavior meant I should leave it as is. Does that seem like the right move to you as well?
- Even without changing the existing ref workaround, this has a slight possibility of having unexpected effects on ConfigurationGroup, if a widget that uses ConfigurationGroup has behavior that depends on the current only-set-up-once behavior of onChange. Do you think that risk warrants only using a ref workaround in PushdownConfig rather than generalizing it to all ConfigurationGroup instances the way I've done here?

### DIFF
--- a/app/cdap/components/ConfigurationGroup/index.tsx
+++ b/app/cdap/components/ConfigurationGroup/index.tsx
@@ -19,7 +19,7 @@ import PropTypes from 'prop-types';
 import withStyles, { WithStyles, StyleRules } from '@material-ui/core/styles/withStyles';
 import { IWidgetJson, PluginProperties } from './types';
 import { processConfigurationGroups, removeFilteredProperties } from './utilities';
-import { objectQuery } from 'services/helpers';
+import { objectQuery, useOnUnmount } from 'services/helpers';
 import defaults from 'lodash/defaults';
 import If from 'components/If';
 import PropertyRow from './PropertyRow';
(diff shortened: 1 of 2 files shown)


app/cdap/services/helpers.js
```diff
@@ -723,6 +723,19 @@ function santizeStringForHTMLID(str) {
   return str.replace(/[ \/]/g, '-');
 }
 
+/**
+ * Wrapper for useEffect that ensures the latest version of the callback will be used, unlike useEffect
+ */
+function useOnUnmount(unmountHandler) {
```
seanfeld, line 729: can we add this to a different file? this helpers file is already very bloated, maybe a custom hooks folder that is more top level? or `services/hooks/useOnUnmount.ts` ? In the spirit of starting to organize things I think it would make sense to keep this separate so its easier to find.

app/cdap/components/ConfigurationGroup/index.tsx
```diff
@@ -169,6 +157,16 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
 
+  // This onUnMount is to make sure we clear out all properties that are hidden.
+  useOnUnmount(() => {
+    const newValues = referenceValueForUnMount.current.values;
+    const configGroups = referenceValueForUnMount.current.configurationGroups;
+    const updatedFilteredValues = removeFilteredProperties(newValues, configGroups);
+    changeParentHandler(updatedFilteredValues);
+  });
+
+  React.useEffect(getOrphanedErrors, [errors]);
```
seanfeld, line 168: Why does it think these lines are changed?

seanfeld (approved):
I'm approving because my comments are more nits but I think we should add custom hooks to separate folder.

app/cdap/components/ConfigurationGroup/index.tsx
lahwran: I moved changeParentHandler to be above useOnUnmount

app/cdap/services/helpers.js
lahwran: To any future code archeologists: we discussed this and decided to split the file after 6.5 release

PR #94, opened 2021-07-12
[CDAP-18005] Re-enable Wrangler E2E tests to work with connection management

njbriggs:
Jira: https://cdap.atlassian.net/browse/CDAP-18005

### DIFF
--- a/cypress/support/constants.ts
+++ b/cypress/support/constants.ts
@@ -22,7 +22,9 @@ const DEFAULT_GCP_SERVICEACCOUNT_PATH = Cypress.env('gcp_service_account_path');
 const DEFAULT_GCP_PROJECTID = Cypress.env('gcp_projectid');
 
 const DEFAULT_GCS_CONNECTION_NAME = `gcs_${getRandomArbitrary()}`;
-const DEFAULT_GCS_FOLDER = 'cdap-gcp-ui-test';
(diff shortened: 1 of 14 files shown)


cypress/support/constants.ts
```diff
@@ -22,7 +22,7 @@ const DEFAULT_GCP_SERVICEACCOUNT_PATH = Cypress.env('gcp_service_account_path');
 const DEFAULT_GCP_PROJECTID = Cypress.env('gcp_projectid');
 
 const DEFAULT_GCS_CONNECTION_NAME = `gcs_${getRandomArbitrary()}`;
-const DEFAULT_GCS_FOLDER = 'cdap-gcp-ui-test';
+const DEFAULT_GCS_FOLDER = '000cdap-gcp-ui-test';
```
seanfeld, line 27: Why 000?

seanfeld (commented):
Looks good but this is big maybe it should get another pair of eyes?

cypress/support/constants.ts
njbriggs: Oh let me add a comment. The test bucket has been pushed down by other buckets so it doesn't appear in the listing.

lahwran (approved):
Looks good to my eye, so I'll approve.

PR #95, opened 2021-07-12
CDAP-18193 - Adds optional prop to errorBanner and Alert components to make page interactable when used as a toast.

seanfeld:
### Bugfix
[jira ticket](https://cdap.atlassian.net/browse/CDAP-18193)

**Code changes**
This adds an optional prop to `Alert` and `Error Banner` components so that the page is interactable when an error message shows up. 

**Context**
This is a short term fix. Modals are normally components that take away all focus from the rest of the page, going into a different mode. We're currently using the reactstrap/modal component instead of a toast, so you can't interact with the page while the error message is there. If we want the error message to be a toast (like this scenario) we should use the toast component that reactstrap has in it or make another toast component that fits our needs.

### DIFF
--- a/app/cdap/components/Alert/Alert.scss
+++ b/app/cdap/components/Alert/Alert.scss
@@ -15,6 +15,9 @@
  */
 @import "../../styles/variables.scss";
 $close-btn-size: 13px;
+.modal.can-edit-page-while-open {
+  height: auto;
+}
 
 .global-alert {
   &.modal-dialog {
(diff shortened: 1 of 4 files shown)


lahwran (approved):
Code change wise, this looks good to me, but I'd like to get signoff from someone on the UX effects. +1 for comment explaining that there is a postponed change. What's the reason to postpone using a toast? is it complicated?

seanfeld:
@lahwran the reason i had to postpone using a toast was development time, unsure about styling changes that would need to get approval and minimizing side effects before 6.5 release.

lahwran:
Update: I ran into this myself while testing https://github.com/cdapio/cdap-ui/pull/98. Looking forward to merge!

PR #128, opened 2021-07-20
CDAP 18260 - fixes background color of add connection button

seanfeld:
### BUGFIX
[CDAP-18260](https://cdap.atlassian.net/browse/CDAP-18260)

changes css to fix the background color of the button

Screen Shot 2021-07-20 at 2 46 27 PM

### DIFF
--- a/app/cdap/components/Connections/AddConnectionBtnModal/index.tsx
+++ b/app/cdap/components/Connections/AddConnectionBtnModal/index.tsx
@@ -17,9 +17,20 @@
 import React, { useState } from 'react';
 import Button from '@material-ui/core/Button';
 import CreateConnectionModal from 'components/Connections/CreateConnectionModal';
+import { makeStyles, Theme } from '@material-ui/core/styles';
+import clsx from 'clsx';
+
+const useStyle = makeStyles<Theme>((theme) => {
+  return {
+    root: {
+      backgroundColor: 'white',
+    },
+  };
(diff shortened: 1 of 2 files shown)


lahwran (approved):

app/cdap/components/Connections/AddConnectionBtnModal/index.tsx
```diff
@@ -35,9 +46,9 @@ export default function AddConnectionBtnModal({ onCreate, className = null }) {
   return (
     <>
       <Button
-        variant="contained"
+        variant="outlined"
         color="primary"
-        className={className}
+        className={clsx(className, classes.root)}
```
njbriggs, line 51: Is setting the style necessary? An outlined primary button should have a white background already: https://material-ui.com/components/buttons/#outlined-buttons

In general, we don't want to custom-code buttons. We want to get to a small number of pre-defined buttons.

seanfeld: @njbriggs the background color is set in the other add connection button and without it looks like this (no background color)
Screen Shot 2021-07-20 at 3 05 37 PM

njbriggs: Ack, you're right.

njbriggs (approved):

PR #139, opened 2021-07-22
[CDAP-18093] Redirect to default connection

lahwran:
Only if:
- connections are browseable (!hideSidePanel)
- no connection provided (!connectionId)

### DIFF
--- a/app/cdap/components/Connections/index.tsx
+++ b/app/cdap/components/Connections/index.tsx
@@ -20,7 +20,8 @@ import Helmet from 'react-helmet';
 import T from 'i18n-react';
 import { Theme } from 'services/ThemeHelper';
 import { ConnectionRoutes } from 'components/Connections/Routes';
-import { MemoryRouter } from 'react-router';
+import { MemoryRouter, Redirect } from 'react-router';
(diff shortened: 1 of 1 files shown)


app/cdap/components/Connections/index.tsx
```diff
@@ -41,6 +41,30 @@ const useStyle = makeStyle(() => {
...
+  // useRef instead of useState to avoid risk of feedback loop
+  const redirected = useRef(false);
+
+  if (
+    // nothing provided by parent of <Connections/>
+    !connectionId &&
+    // but we have a default
+    initialConnectionId &&
+    // and we're not at the default
+    (params as any).connectionid !== initialConnectionId &&
```
njbriggs: If I follow a link like `http://localhost:11011/cdap/ns/default/connections/connection-name-here`, will this redirect to the default?

lahwran: Good call, that should have been !params.connectionid, except it shouldn't have been that as it wasn't in a route that would gate it correctly. See new review.

njbriggs (approved):

PR #148, opened 2021-07-29
CDAP 18283 - Add plugin name to use as reference name default

seanfeld:
### Bugfix

Use plugin.name as the default referenceName for bigquery browse inside of studio. Was unsure of what to use for the name - it looks like we could parse the name of the big query connection from inside the connection string like below:
Screen Shot 2021-07-28 at 8 07 06 PM
But there was also name which defaults to `BigQueryTable`.

https://cdap.atlassian.net/browse/CDAP-18283

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
@@ -14,7 +14,7 @@
  * the License.
  */
 
-import * as React from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { isNilOrEmptyString } from 'services/helpers';
 import {
   exploreConnection,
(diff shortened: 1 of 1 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -166,6 +166,7 @@ export function GenericBrowser({ selectedConnection }) {
       const plugin = spec?.relatedPlugins?.[0];
 
       const properties = plugin?.properties;
+      properties.referenceName = plugin.name;
```
lahwran: - Correctness: The other accesses to `plugin` around this line are using optional chaining, so you likely also need to use optional chaining, and properties might be null. I'm surprised and a bit disappointed in typescript that it didn't infer this and warn you, actually. (Perhaps it simply believes the optional chaining is unnecessary in the first place, though, haha.)
- Question: Can this ever overwrite a user's input of referenceName? is GenericBrowser ever used to navigate configured nodes in a pipeline, for example?

lahwran (approved):

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -164,10 +164,14 @@ export function GenericBrowser({ selectedConnection }) {
     try {
       const spec = await getPluginSpec(entity, selectedConnection);
       const plugin = spec?.relatedPlugins?.[0];
-
       const properties = plugin?.properties;
       const schema = plugin?.schema;
 
+      // if reference name is already set, don't set default.
+      if (!properties.referenceName) {
+        properties.referenceName = plugin?.name;
```
yaojiefeng, line 172: I don't think the change made here is correct, what the JIRA mentioned is to fill in the missing connection name in the connection widget, not the reference name. @njbriggs can you help take a look here?

PR #149, opened 2021-07-29
[CDAP-17629] Add special case null check for missing field

lahwran:


app/hydrator/services/hydrator-upgrade-service.js
```diff
@@ -174,6 +174,12 @@ class HydratorUpgradeService {
 
   _checkErrorStages(stages, pluginsMap) {
     let transformedStages = [];
+    if (!stages || !stages.forEach) {
```
seanfeld, line 177: is this meant to check if its a immutable map as well as an array?

lahwran: This is meant to guard against unknown types since I only have a single failure example and I don't know what's possible. I could clarify the comment if you like!

seanfeld (approved):
nit: `.forEach` check is weird to me - can we be more explicit?

app/hydrator/services/hydrator-upgrade-service.js
seanfeld: yeah can we add a comment then? or is that linked in the jira?

PR #150, opened 2021-07-29
cherry pick 18283 - add plugin name to use as reference name default

seanfeld:
https://cdap.atlassian.net/browse/CDAP-18283

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
@@ -14,7 +14,7 @@
  * the License.
  */
 
-import * as React from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { isNilOrEmptyString } from 'services/helpers';
 import {
   exploreConnection,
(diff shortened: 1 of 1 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -164,10 +164,14 @@ export function GenericBrowser({ selectedConnection }) {
     try {
       const spec = await getPluginSpec(entity, selectedConnection);
       const plugin = spec?.relatedPlugins?.[0];
-
       const properties = plugin?.properties;
       const schema = plugin?.schema;
 
+      // if reference name is already set, don't set default.
+      if (!properties.referenceName) {
+        properties.referenceName = plugin?.name;
```
yaojiefeng, line 172: I don't think the change made here is correct, what the JIRA mentioned is to fill in the missing connection name in the connection widget, not the reference name. @njbriggs can you help take a look here?

seanfeld: talked via chat - going to make another pr and abandon this one

PR #156, opened 2021-07-31
[PLUGIN-769] Added dropdown menu option to call the explode record directive

sechegaray:


app/cdap/components/DataPrep/Directives/Explode/index.js
```diff
@@ -92,6 +93,11 @@ export default class Explode extends Component {
     this.execute([directive]);
   }
 
+  explodeRecordByFlattening() {
+    let directive = `explode-record :${this.props.column.toString()}`;
```
seanfeld: should be const

seanfeld: i realize you were just following the pattern that existed elsewhere on this file but that pattern is wrong :)

seanfeld (changes requested):

app/cdap/components/DataPrep/Directives/Explode/index.js
sechegaray: fixed

seanfeld (approved):

PR #159, opened 2021-08-03
[CDAP-18295] Make use of default connection an explicit flag

lahwran:


app/cdap/components/DataPrepHome/index.js
```diff
@@ -300,6 +300,7 @@ export default class DataPrepHome extends Component {
           <Connections
             mode={mode || 'ROUTED_WORKSPACE'}
             onWorkspaceCreate={mode === 'INMEMORY' ? null : this.onWorkspaceCreate}
+            allowDefaultConnection={true}
```
lahwran, line 303: @njbriggs is this correct? Should DataPrepHome use the default connection?

app/cdap/components/Connections/index.tsx
```diff
@@ -73,6 +73,7 @@ export default function Connections({
   initPath,
   connectorType,
   hideAddConnection,
+  allowDefaultConnection = true,
```
lahwran, line 76: This has a default because my editor yelled at me in a way I didn't feel like debugging when I tried to use `render={() => <Connections allowDefaultConnection={true}/>}` in the `<Route/>` in https://github.com/cdapio/cdap-ui/blob/ee5f2e1b040c099394586b76df707557afe6d00c/app/cdap/components/Home/index.js#L150 - however, I think it might make more sense to make that change and have `allowDefaultConnection` default to something else, eg `false`, or `undefined`, or just require `allowDefaultConnection`. Reviewer, please advise :)

app/cdap/components/DataPrepHome/index.js
njbriggs: Yes. `DataPrep` is Wrangler

app/cdap/components/Connections/index.tsx
njbriggs: Could you set the default in `ConnectionsContext`? That might be a little cleaner. But I think setting it this way, not in the `Route`, is the way to go.

njbriggs (approved):

PR #165, opened 2021-08-06
CDAP-18263 | disable link if non editable profiles

seanfeld:
### Bugfix
Disables link if there are no editable profiles. 

https://cdap.atlassian.net/browse/CDAP-18263


app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
```diff
@@ -94,6 +95,7 @@ export default class ProfileCustomizeContent extends PureComponent {
   };
 
   render() {
+    const editablePropertiesFromProfile = this.props.editablePropertiesFromProfile;
```
lahwran, line 98: low-value nit: `const {editablePropertiesFromProfile} = this.props;`

lahwran (approved):
no changes needed that I see

PR #168, opened 2021-08-10
CDAP - 18324 | moves getError to api handler file and wraps create workspace with it

seanfeld:
### Bugfix

https://cdap.atlassian.net/browse/CDAP-18324

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/apiHelpers.ts
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/apiHelpers.ts
@@ -57,7 +57,7 @@ export function createWorkspace({ entity, connection, limit = 1000 }) {
   )
     .toPromise()
     .catch((e) => {
-      return getApiErrorMessage(e);
+      throw getApiErrorMessage(e);
     });
 }
 
--- a/app/cdap/components/Connections/Browser/GenericBrowser/apiHelpers.ts
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/apiHelpers.ts
@@ -35,6 +35,10 @@ export function exploreConnection({ connectionid, path = '/' }) {
   ).toPromise();
 }
 
+export function getApiErrorMessage(e) {
+  return e?.message || e?.response?.message || e?.statusCode || 'Unknown error';
+}
+
 export function createWorkspace({ entity, connection, limit = 1000 }) {
   const { path } = entity;
   const body = {
@@ -50,7 +54,11 @@ export function createWorkspace({ entity, connection, limit = 1000 }) {
       context: getCurrentNamespace(),
     },
     body
-  ).toPromise();
+  )
+    .toPromise()
+    .catch((e) => {
+      return getApiErrorMessage(e);
+    });
 }
 
 export function getPluginSpec(entity, connection) {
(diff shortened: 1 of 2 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/apiHelpers.ts
```diff
@@ -50,7 +54,11 @@ export function createWorkspace({ entity, connection, limit = 1000 }) {
       context: getCurrentNamespace(),
     },
     body
-  ).toPromise();
+  )
+    .toPromise()
+    .catch((e) => {
+      return getApiErrorMessage(e);
```
njbriggs: I'm rusty on my Promises. Will returning the string from `.catch` result in a rejected Promise or a resolved Promise?

seanfeld: my understanding is it will catch the error from a rejected promise, then return it so that it can hit the catch in try catch. @lahwran do you remember? I feel a bit rusty as well.

njbriggs: Safer to just `return Promise.reject(getApiErrorMessage(e));`?

seanfeld: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

seanfeld: my intent was to not have to call `getErrorMessage` on each individual `catch(e) {}` block for logging purposes. I think i remembered it working this way and from a small initial test it seemed to be but i'm not 100% sure.

lahwran: Nick is correct, catch's return value becomes a resolved promise, not a rejected one. if you'd like to reject, raise a value as an error, or do what nick suggested.
Screen Shot 2021-08-10 at 1 13 00 PM

app/cdap/components/Connections/Browser/GenericBrowser/apiHelpers.ts
```diff
@@ -35,6 +35,10 @@ export function exploreConnection({ connectionid, path = '/' }) {
   ).toPromise();
 }
 
+export function getApiErrorMessage(e) {
+  return e?.message || e?.response?.message || e?.statusCode || 'Unknown error';
```
lahwran, line 39: should this now check if `e` is a string and if so, return it as is, to ensure `getApiErrorMessage(getApiErrorMessage(e)) == getApiErrorMessage(e)`? this may help the other comment

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -172,7 +169,7 @@ export function GenericBrowser({ selectedConnection }) {
         schema,
       });
     } catch (e) {
-      setError(`Failed to get plugin information. Error: ${getErrorMessage(e)}`);
+      setError(`Failed to get plugin information. Error: ${getApiErrorMessage(e)}`);
```
lahwran, line 172: does this e come from a different place? getPluginSpec is an api function that calls into the code you modified in apiHelpers.ts, right?

lahwran: Ah I see that I misread how general the modification in apihelpers was

lahwran (approved):

njbriggs (approved):

PR #171, opened 2021-08-11
CDAP 18339 - check if use connection is true before using connection name in plugin connection browser

seanfeld:
# Bugfix

### Changes
- plugin connection browser checks if use connection is turned on to use connection name to hide sidebar of connection modal
- Converted AbstractWidget to functional component

### Buglink
https://cdap.atlassian.net/browse/CDAP-18339
-https://cdap.atlassian.net/browse/CDAP-18338-

### DIFF
--- a/app/cdap/components/AbstractWidget/ToggleSwitchWidget/index.tsx
+++ b/app/cdap/components/AbstractWidget/ToggleSwitchWidget/index.tsx
@@ -52,8 +52,6 @@ const ToggleSwitchWidgetView: React.FC<IToggleToggleSwitchProps> = ({
   disabled,
   classes,
   dataCy,
-  updateAllProperties,
-  extraConfig,
 }) => {
   const onValue = objectQuery(widgetProps, 'on', 'value') || 'on';
   const offValue = objectQuery(widgetProps, 'off', 'value') || 'off';
@@ -62,20 +60,7 @@ const ToggleSwitchWidgetView: React.FC<IToggleToggleSwitchProps> = ({
   const isOn = value === onValue;
 
   function toggleSwitch() {
-    // if connection toggle
-    if (isOn && extraConfig.properties.hasOwnProperty('useConnection')) {
-      updateAllProperties(
-        {
-          connection: '',
-          useConnection: false,
-          serviceAccountType: 'filePath',
-          serviceFilePath: 'auto-detect',
-        },
-        { updateFilteredConfigurationGroups: true }
-      );
-    } else {
-      onChange(isOn ? offValue : onValue);
-    }
+    onChange(isOn ? offValue : onValue);
   }
 
   return (
--- a/app/cdap/components/AbstractWidget/ToggleSwitchWidget/index.tsx
+++ b/app/cdap/components/AbstractWidget/ToggleSwitchWidget/index.tsx
@@ -52,6 +52,8 @@ const ToggleSwitchWidgetView: React.FC<IToggleToggleSwitchProps> = ({
   disabled,
   classes,
   dataCy,
+  updateAllProperties,
+  extraConfig,
 }) => {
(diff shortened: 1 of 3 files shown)


app/cdap/components/AbstractWidget/index.tsx
```diff
@@ -59,29 +59,37 @@ export interface IWidgetProps<T = any> {
 }
 
 interface IAbstractWidgetProps extends IWidgetProps {
-  type: string;
+  type?: string;
```
njbriggs, line 62: When is this optional?

app/cdap/components/AbstractWidget/index.tsx
```diff
@@ -44,7 +44,7 @@ export interface IWidgetExtraConfig {
 
 export interface IWidgetProps<T = any> {
   widgetProps?: T;
-  value: string | number;
+  value?: string | number;
```
njbriggs, line 47: This is not set when the field is empty?

app/cdap/components/AbstractWidget/index.tsx
seanfeld: there were a ton of type errors on unit tests without this

app/cdap/components/AbstractWidget/index.tsx
seanfeld: this one might've been a mistake let me double check it

app/cdap/components/AbstractWidget/index.tsx
seanfeld: actually nvm, the errors were in stories
```
[3] ERROR in /Users/podesta/goog/cdap-ui/app/cdap/components/AbstractWidget/Widgets.stories.tsx(81,38):
[3] TS2741: Property 'value' is missing in type '{ type: string; }' but required in type 'IAbstractWidgetProps'.
[3] ERROR in /Users/podesta/goog/cdap-ui/app/cdap/components/AbstractWidget/Widgets.stories.tsx(83,27):
[3] TS2741: Property 'value' is missing in type '{ type: string; }' but required in type 'IAbstractWidgetProps'.
[3] ERROR in /Users/podesta/goog/cdap-ui/app/cdap/components/AbstractWidget/Widgets.stories.tsx(86,4):
[3] TS2741: Property 'value' is missing in type '{ type: string; widgetProps: { dropdownOptions: { value: string; label: string; }[]; }; }' but required in type 'IAbstractWidgetProps'
```

seanfeld: whoops i got these mixed up haha

app/cdap/components/AbstractWidget/ToggleSwitchWidget/index.tsx
```diff
@@ -60,8 +62,22 @@ const ToggleSwitchWidgetView: React.FC<IToggleToggleSwitchProps> = ({
   const isOn = value === onValue;
 
   function toggleSwitch() {
-    onChange(isOn ? offValue : onValue);
+    // if connection toggle
+    if (isOn && extraConfig.properties.hasOwnProperty('useConnection')) {
+      updateAllProperties(
```
njbriggs: I don't think we want to do this. It's GCP-specific - it adds these fields for all sources, even if they don't use them. And what if we want to set defaults for other types, like AWS, or databases (once we support connections there again)?

It also ties this widget to other fields in a way we've tried to avoid so far.

The only way I can think of to handle this now would be in `ConfigurationGroup` in `handleValueChanges`: if any fields are newly shown, have no value, and have a default, set the default. If that seems like too much work, then we should re-prioritize.

app/cdap/components/AbstractWidget/index.tsx
seanfeld: `type?` is necessary because default widget props doesn't contain it.

njbriggs (approved):

PR #177, opened 2021-08-13
Allows spaces in connection name 

itsankit-google:
[CDAP-18346] Allows spaces in connection name 

[CDAP-18346]: https://cdap.atlassian.net/browse/CDAP-18346


lahwran (approved):
Looks good to me, just need to test that using connections with spaces in the BQ source works correctly

PR #179, opened 2021-08-14
[CDAP-18326] Restore edit mode as a read-only view

lahwran:


seanfeld (approved):

app/cdap/components/NamespaceAdmin/Connections.tsx
```diff
@@ -153,6 +159,10 @@ const ConnectionsView: React.FC<IConnectionsProps> = ({ connections }) => {
         <TableBody>
           {connections.map((conn) => {
             const actions: IAction[] = [
+              {
+                label: 'Edit',
+                actionFn: () => editConnection(conn),
```
njbriggs: We don't want edit here, do we?

lahwran: Whoops, good call

njbriggs (approved):

PR #188, opened 2021-08-17
[CDAP-18383] Don't attempt default conn in namespace

lahwran:
Also fixes error that was occurring when opening the connections page with no default connection, due to the way the custom path splitting logic was pulling off "/connections" as though it was a connection name. Using react-router's path parsing fixes that, and retains the lock-step updating needed to reduce spurious useEffects.

### DIFF
--- a/app/cdap/components/Connections/index.tsx
+++ b/app/cdap/components/Connections/index.tsx
@@ -121,7 +121,7 @@ export default function Connections({
         const disabledTypes = config?.connectionConfig?.disabledTypes;
         const def = config?.connectionConfig?.defaultConnection;
 
-        if (def && allowDefaultConnection && !connectionId) {
+        if (def && getCurrentNamespace() === 'default' && allowDefaultConnection && !connectionId) {
           updatedState = {
             ...updatedState,
             connectionId: def,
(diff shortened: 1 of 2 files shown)


app/cdap/components/Connections/index.tsx
```diff
@@ -121,7 +121,7 @@ export default function Connections({
         const disabledTypes = config?.connectionConfig?.disabledTypes;
         const def = config?.connectionConfig?.defaultConnection;
 
-        if (def && allowDefaultConnection && !connectionId) {
+        if (def && getCurrentNamespace() === 'default' && allowDefaultConnection && !connectionId) {
```
njbriggs, line 124: Is it possible that the default connection is in a different namespace?

njbriggs: That is, say the default connection exists in namespace `foo` and the user is currently in namespace `bar` (or `default`). Is that possible? If so, it seems like we have to find a place to check for the existence of the namespace/connection combo.

lahwran: Yaojie said it is not currently possible. Only the default namespace can have a default connection right now.

njbriggs (approved):

PR #193, opened 2021-08-25
Removes sourcemap from production cdap.js build

antony-britto:
Removes sourcemap from production cdap.js build

'eval-source-map' (This bloats final JS by adding sourceMap into the bundle, ex: below)
https://vitalii-test-rbac-rc6-vitalii-playground-294422-dot-r.datafusion-dev.googleusercontent.com/cdap_assets/cdap.dd3bbd7e2a67cbb81646.js?c4a6b45056018b32011c

### DIFF
--- a/webpack.config.cdap.js
+++ b/webpack.config.cdap.js
@@ -217,7 +217,6 @@ if (mode === 'development') {
 
 var webpackConfig = {
   mode: isModeProduction(mode) ? 'production' : 'development',
-  devtool: 'eval-source-map',
   context: __dirname + '/app/cdap',
   entry: {
     cdap: ['@babel/polyfill', './cdap.js'],


webpack.config.cdap.js
```diff
@@ -217,7 +217,6 @@ if (mode === 'development') {
 
 var webpackConfig = {
   mode: isModeProduction(mode) ? 'production' : 'development',
-  devtool: 'eval-source-map',
```
seanfeld, line 220: shouldn't this copy the line above? `isModeProduction(mode) ? '' : 'eval-source-map`?

antony-britto: https://github.com/cdapio/cdap-ui/blob/develop/webpack.config.cdap.js#L260, handles dev mode. devtool documentation has no mention about '' value, hence removed it.

seanfeld (approved):

PR #198, opened 2021-09-06
Adds CMEK revocation error page

antony-britto:


njbriggs:
Can you create a Jira for this?

Also, how will it be triggered? How will it work for an active session if the key is revoked/disabled (i.e., some error comes over the websocket)?

antony-britto:
Created jira for the same https://cdap.atlassian.net/browse/CDAP-18527.

Added a validation in the web-socket layer to redirect to the error page.

app/cdap/services/datasource/index.js
```diff
@@ -54,7 +55,7 @@ export default class Datasource {
       }
 
       genericResponseHandlers.forEach((handler) => handler(data));
-
+      redirectOnError(data.response);
```
njbriggs: I don't think we should redirect on this error. If the user's in the middle of something, they'll lose their work, which could be recovered if the error is transient or the downtime is short. Also, some of our pages (like Pipeline Studio) have an `onBeforeUnload` handler which could trigger, making this even more confusing for the user.

The next block handles errors. It may not be perfect, but we should build on that.

antony-britto: Converted this into an error dialog which gets triggered on API errors. It's used in both angular & react app.

njbriggs (approved):
Thanks for making the change!

PR #201, opened 2021-09-13
Adds validation for connection name

antony-britto:


app/cdap/components/Connections/Create/index.tsx
```diff
@@ -135,6 +135,9 @@ export function CreateConnection({
     if (name.trim() === '') {
       setError('Connection name must not be empty.');
       return;
+    } else if (!new RegExp('^[\\w\\s-]+$').test(name)) {
+      setError('Connection name must not contain special characters.');
```
njbriggs: The message should be prescriptive about what is allowed (`Connection names can only include letters, numbers, and underscore`, etc.)

antony-britto: Done

app/cdap/components/Connections/Create/index.tsx
```diff
@@ -135,6 +135,9 @@ export function CreateConnection({
     if (name.trim() === '') {
       setError('Connection name must not be empty.');
       return;
+    } else if (!new RegExp('^[\\w\\s-]+$').test(name)) {
+      setError('Connection name can only include alphabets, numbers, hypen and underscore.');
```
njbriggs: Nit: `letters` instead of `alphabets`

njbriggs (approved):
LGTM, just a slight tweak to the message

PR #207, opened 2021-09-16
Adds metadata home page in react app

antony-britto:
1) Add top level routing for metadata pages
2) Adds metadata home page component
3) Adds a feature flag and handles redirection between angular & react pages.

### DIFF
--- a/app/cdap/components/Metadata/Home/index.tsx
+++ b/app/cdap/components/Metadata/Home/index.tsx
@@ -1,0 +1,134 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import React, { useState } from 'react';
+import T from 'i18n-react';
+import InputLabel from '@material-ui/core/InputLabel';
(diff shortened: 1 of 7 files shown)


app/cdap/components/Metadata/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  function checkFeatureFlag(componentToRender: JSX.Element, urlToRedirect: string) {
+    if (!Theme.isMetadataInReact) {
+      window.location.href = urlToRedirect;
+    } else {
+      return componentToRender;
+    }
+  }
+
+  return (
+    <React.Fragment>
```
seanfeld: Lets just use the `<></>` shorthand for react fragments.

app/cdap/services/ThemeHelper.ts
```diff
@@ -223,6 +224,7 @@ interface IThemeObj {
   featureNames?: IFeatureNames;
   showSqlPipeline?: boolean;
   showCDC?: boolean;
+  isMetadataInReact?: boolean;
```
seanfeld, line 227: Can we add a todo here? Eventually we want the feature flags to be in a context provider situation and I don't think keeping it in themes makes sense long term.

app/cdap/components/Metadata/Home/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+const MetadataHome: React.FC = () => {
+  const searchQueryRegex = new RegExp('^(?![*]).*$');
+  const [error, setError] = useState(false);
+
+  const handleSearch = (event) => {
+    const searchQuery = event.target.value;
+    const isValid = searchQueryRegex.test(searchQuery);
+    setError(!isValid);
+    if (event.keyCode === 13 && searchQuery.trim() !== '' && isValid) {
```
seanfeld: I think we can do `event.key === 'Enter'` or `event.code === 'Enter'` - i think its much easier/clearer than keycode.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

antony-britto: Done

seanfeld (approved):

PR #208, opened 2021-09-16
Update dependency

lahwran:
It wasn't installing on my machine until I did "yarn add babel-plugin-styled-components". This is the result of that operation.


seanfeld:
@lahwran do we still need this?

lahwran:
I don't think so, because I had forgotten about it and my localdev is working anyway.

PR #211, opened 2021-09-27
Bugfix - use the shorthand for the array constructor instead of array.apply

seanfeld:
## Bugfix
Array.apply() over a certain size creates a max callstack range error. Array constructor doesn't have that problem (at least up to a certain value).

https://cdap.atlassian.net/browse/PLUGIN-917

### DIFF
--- a/app/cdap/components/PipelineSummary/RunsGraphHelpers.js
+++ b/app/cdap/components/PipelineSummary/RunsGraphHelpers.js
@@ -178,7 +178,7 @@ export function getGapFilledAccumulatedData(data, numOfDataPoints) {
   let maxx = data[data.length - 1].x;
   let numberOfEntries = maxx - minx;
   let lasty = miny;
-  let finalData = Array.apply(null, { length: numberOfEntries + 1 }).map((i, index) => {
+  let finalData = Array(numberOfEntries + 1).map((i, index) => {
     let matchInActualData = data.find((d) => d.x === minx + index);
     if (!isNil(matchInActualData)) {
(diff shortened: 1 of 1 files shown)


app/cdap/components/PipelineSummary/RunsGraphHelpers.js
```diff
@@ -178,7 +178,7 @@ export function getGapFilledAccumulatedData(data, numOfDataPoints) {
   let maxx = data[data.length - 1].x;
   let numberOfEntries = maxx - minx;
   let lasty = miny;
-  let finalData = Array.apply(null, { length: numberOfEntries + 1 }).map((i, index) => {
+  let finalData = Array(numberOfEntries + 1).map((i, index) => {
```
njbriggs, line 181: Did you verify that this fixes the issue? The syntax here is not exactly the same as what you discussed (`foo.apply(null, new Array(n)]` vs `Array.apply(null, { length: n })` though if `{ length: n }` is treated as array-like, it might show the same behavior).

seanfeld: {length: n} is treated as an empty array (array-like as you said).
Screen Shot 2021-09-27 at 11 17 55 PM

```
Array.apply(null, {length: 25000000});
>> VM486:1 Uncaught RangeError: Maximum call stack size exceeded
    at <anonymous>:1:7
(anonymous) @ VM486:1
Array(25000000);
>> (25000000) [empty × 25000000]

njbriggs (approved):

PR #214, opened 2021-09-30
Adds metadata search results page in react app

antony-britto:


app/cdap/components/Metadata/SearchBar/index.tsx
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
```
seanfeld: @njbriggs is this what we're still putting above files?

app/cdap/components/Metadata/SearchBar/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { getCurrentNamespace } from 'services/NamespaceStore';
+import Paper from '@material-ui/core/Paper';
+import InputBase from '@material-ui/core/InputBase';
+import IconButton from '@material-ui/core/IconButton';
+import ResourceCenterButton from 'components/ResourceCenterButton';
+
+const I18N_PREFIX = 'features.MetadataSearch';
+
+const Container = styled.div`
+  background: #dfdfdf;
```
seanfeld: Hmm we need to start thinking about theming or reusing these colors etc @njbriggs @lahwran

app/cdap/components/Metadata/SearchBar/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  flex: 1;
+  margin-left: 8px;
+  font-size: 13px;
+`;
+
+const ResourceAddButton = styled(ResourceCenterButton)`
+  top: 80px;
+`;
+
+interface ISearchBar {
```
seanfeld: We should probably include `Props` in the interface name right? @lahwran @njbriggs

app/cdap/components/Metadata/SearchResults/ResultList/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+`;
+
+const Description = styled.p`
+  color: #999;
+  margin: 0;
+  padding: 0;
+`;
+
+export interface IResultList {
```
seanfeld: `Props` ? Also is this used in another component? If not we probably shouldn't export it

app/cdap/components/Metadata/SearchResults/helper.ts
```diff
@@ -0,0 +1,263 @@
...
+  metadataFiltersList: IFilterItem[]
+): IParsedResult {
+  let count = 0;
+  // Removing special characters from search query
+  const replaceRegex = new RegExp('[^a-zA-Z0-9_-]', 'g');
+  const searchTerm = searchQuery.replace(replaceRegex, '');
+  const regex = new RegExp(searchTerm, 'ig');
+  const results = entities.map((entityObj) => {
+    const result = {
+      name: entityObj.entity.details.dataset,
```
seanfeld: are these all guaranteed to exist?

app/cdap/components/Metadata/SearchResults/helper.ts
```diff
@@ -0,0 +1,263 @@
+/*
+ * Copyright © 2017 Cask Data, Inc.
```
seanfeld: All the copyrights have different years on them - https://cdap.atlassian.net/wiki/spaces/CE/pages/1594100059/Setting+up+Environment#SettingUpEnvironment-auto-copyright-header-intellij found the copyright header we should be using

app/cdap/components/Metadata/SearchResults/helper.ts
```diff
@@ -0,0 +1,263 @@
...
+  'Name',
+  'Description',
+  'User tags',
+  'System tags',
+  'User properties',
+  'System properties',
+  'Schema',
+];
+
+export function parseResult(
```
seanfeld, line 85: Should we add tests for some of these? Also jsdoc them?

app/cdap/components/Metadata/SearchResults/helper.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  return entitySearchResults.filter((result) => {
+    if (result.queryFound.length === 0) {
+      return true;
+    }
+    return intersection(metadataFilter, result.queryFound).length > 0;
+  });
+}
+
+export function applyFilter(
```
seanfeld, line 256: jsdoc - tests?

app/cdap/components/Metadata/SearchResults/helper.ts
```diff
@@ -0,0 +1,263 @@
...
+    default:
+  }
+  return {
+    results: filterResults(results, entityFilters, metadataFilters),
+    entityFilters,
+    metadataFilters,
+  };
+}
+
+export function getResultCount(currentPage: number, total: number) {
```
seanfeld: I think we want to jsdoc most exported funcitons right?

seanfeld (changes requested):
Looks great! but we should probably add some test on some of those more complicated functions and we need to change the copyright header years

app/cdap/components/Metadata/SearchBar/index.tsx
njbriggs: All new files should say `2021`

app/cdap/components/Metadata/SearchBar/index.tsx
njbriggs: Let's try to stick to colors in https://github.com/cdapio/cdap-ui/blob/develop/app/cdap/components/ThemeWrapper/colors.ts. `grey[500]` is `#dbdbdb` and might be a close enough match.

app/cdap/components/Metadata/SearchBar/index.tsx
njbriggs: Yes, we usually do.

app/cdap/components/Metadata/SearchBar/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const searchQueryRegex = new RegExp('^(?![*]).*$');
+  const searchPlaceholder = T.translate(`${I18N_PREFIX}.searchPlaceholder`);
+  const [error, setError] = useState(false);
+
+  const handleSearch = (event) => {
+    const searchQuery = event.target.value;
+    const isValid = searchQueryRegex.test(searchQuery);
+    setError(!isValid);
+    if (event.key === 'Enter' && searchQuery.trim() !== '' && isValid) {
+      window.location.href = `/ns/${getCurrentNamespace()}/metadata/search/${searchQuery.trim()}/result`;
```
njbriggs: Nit: Should this fire an `onChange` to its parent and let the parent update the URL?

app/cdap/components/Metadata/SearchBar/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+  };
+
+  return (
+    <>
+      <Container>
+        <SearchForm>
+          <SearchInput
```
njbriggs, line 77: Maybe not this PR, but we should standardize our search input look and behavior. See also the one in the connection browser.

app/cdap/components/Metadata/SearchResults/ResultList/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+            >
+              {result.name}
+            </a>
+          </Title>
+          <Description>
+            <span className={result.icon}></span> {result.type}
+          </Description>
+          <Description>
+            {T.translate(`${I18N_PREFIX}.created`)}{' '}
+            {moment(Number(result.createDate)).format('MMM DD, YYYY')}
```
njbriggs: Can this use `humanReadableDate`?

app/cdap/components/Metadata/SearchBar/index.tsx
njbriggs: https://cdap.atlassian.net/browse/CDAP-18520

app/cdap/components/Metadata/SearchResults/ResultList/index.tsx
antony-britto: No, this shows "Sep 30, 2021" but humanReadableDate shows 09-30-2021

app/cdap/components/Metadata/SearchResults/helper.ts
antony-britto: It's an API response, anyways added optional chaining.

app/cdap/components/Metadata/SearchBar/index.tsx
antony-britto: Done

app/cdap/components/Metadata/SearchBar/index.tsx
antony-britto: Done

app/cdap/components/Metadata/SearchResults/helper.ts
antony-britto: Done

app/cdap/components/Metadata/SearchResults/helper.ts
antony-britto: Done

app/cdap/components/Metadata/SearchResults/helper.ts
seanfeld: @njbriggs does optional chaining still break the build?

app/cdap/components/Metadata/SearchResults/ResultList/index.tsx
seanfeld: moment.js is awesome

njbriggs: It is - but we should be planning to move off of it. https://momentjs.com/docs/#/-project-status/

We should start centralizing our date formatting logic. See https://github.com/cdapio/cdap-ui/blob/develop/app/cdap/services/DataFormatter/index.ts for an example. What is the input date format here?

seanfeld: okay i guess we should start using day.js ? 2kb replacement for moment that is immutable? Is that format service because we have to have generalizable stuff for the abstract widget factory? Thats fine for that but I'd much rather export the individual formatters as well IE `export function formatDateTime`

njbriggs: The format service was intended for use in a data table where each cell's type is determined by some metadata. It totally makes sense to split out specific formatters for a use case like this.

We can look at `day.js` but I wonder what the size it is if you bring in internationalized formats. Maybe use `day.js` or similar for parsing and `Intl.DateTimeFormat` for display.

app/cdap/components/Metadata/SearchResults/helper.ts
njbriggs: Yup. That reminds me, we need to figure out how best to get on a newer Node.

app/cdap/components/Metadata/SearchResults/helper.ts
antony-britto: Done

app/cdap/components/Metadata/SearchResults/ResultList/index.tsx
antony-britto: Added a helper method based on Intl.DateTimeFormat

app/cdap/components/Metadata/SearchBar/index.tsx
antony-britto: Done

seanfeld (approved):

njbriggs:
> Added a helper method based on Intl.DateTimeFormat

This is not really something we support yet because of our use of an old node version for unit testing. If we run into any issues with it, we'll need to revert to `moment`.

That old node version is holding us back in a lot of ways, so I'm going to take a stab at updating it.

PR #216, opened 2021-10-07
Backing page for transformations

seanfeld:
Initial commit for transformations, new select columns page and other improvements
Screen Shot 2021-10-07 at 3 25 34 PM

I've started removing unused imports wherever I see them as well as If components. I moved @lahwran 's useonunmount custom hook to a helper.ts file.  Rewrote the old select columns page to use hooks and styled components. Added the styled components theme wrapper so we have access to the theme colors within styled components.


app/cdap/components/ThemeWrapper/index.tsx
```diff
@@ -28,12 +28,14 @@ import {
...
-  PaletteOptions,
-} from '@material-ui/core/styles/createPalette';
+import { ThemeProvider as StyledThemeProvider } from 'styled-components';
+// this isn't used?
+// import {
+//   PaletteColor,
+//   PaletteColorOptions,
+//   Palette,
+//   PaletteOptions,
+// } from '@material-ui/core/styles/createPalette';
```
lahwran: It's probably safe to delete instead of commenting - it's always in git if we need it

lahwran (approved):
LGTM

app/cdap/components/ThemeWrapper/index.tsx
seanfeld: i commented and left it for this pr review - my plan is to delete it :)

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/SelectColumnsWithTransforms.tsx
```diff
@@ -0,0 +1,549 @@
+/*
```
njbriggs: Since this is a new file, can we split it up into multiple smaller files?

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/SelectColumnsWithTransforms.tsx
```diff
@@ -0,0 +1,549 @@
...
+// const Overlay = styled.div`
+//   background-color: ${(props) => props.theme.palette.white[50]};
+//   opacity: 0.7;
+//   top: 160px;
+//   bottom: 0;
+//   left: 0;
+//   right: 0;
+//   position: absolute;
+//   z-index: 6;
+// `;
```
njbriggs: Remove commented-out block?

app/cdap/components/Replicator/Create/index.tsx
```diff
@@ -22,18 +22,15 @@ import Content from 'components/Replicator/Create/Content';
...
-import { fetchPluginInfo, constructTablesSelection } from 'components/Replicator/utilities';
+import { constructTablesSelection } from 'components/Replicator/utilities';
 
-import { PluginType } from 'components/Replicator/constants';
 import LoadingSVGCentered from 'components/LoadingSVGCentered';
 import uuidV4 from 'uuid/v4';
 import { MyReplicatorApi } from 'api/replicator';
-import { generateTableKey, convertConfigToState } from 'components/Replicator/utilities';
-import { Map, Set, fromJS } from 'immutable';
+import { convertConfigToState } from 'components/Replicator/utilities';
```
njbriggs: Can this be combined with L25?

app/cdap/services/helpers-2.ts
```diff
@@ -0,0 +1,53 @@
+/*
```
njbriggs: Is there a better name than `helpers-2`? `hook-utils` or something?

seanfeld: yeah, these are in a different area now, the main thing is i didn't want to write anymore new javascript so that was a temp file.

app/cdap/components/Replicator/Create/index.tsx
seanfeld: good catch

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/SelectColumnsWithTransforms.tsx
seanfeld: split out the reducer, types and styled components

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
```diff
@@ -16,221 +16,37 @@
...
+  GridWrapper,
+  Header,
+  LoadingContainer,
+  Root,
+  RadioContainer,
+  StyledCheckbox,
+  StyledRadio,
+  StyledRadioGroup,
+  SubtitleContainer,
+} from './styledComponents';
```
lahwran: nit: I might name this "styles" to differentiate it from the upstream styled components

app/cdap/services/react/customHooks/useDebounce.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { useEffect, useState } from 'react';
+
+/**
+ * Custom hook to return a debounced value
+ * @param value - value to return after the delay
+ * @param delay [delay=300] - delay in ms for debounce to wait
+ * @returns debounced value
+ */
+export function useDebounce<T>(value: T, delay = 300): T {
```
lahwran, line 25: You've gone back and forth here, so I don't want to belabor the point; consider this optional. But I generally feel we should probably just have a customHooks.ts file that has our general hooks in it at the moment. My take is that if a file has fewer than 200 lines, it's probably too short, but if it's more than 500 lines, it's probably time to break it up.

My thinking here is that a file is effectively a directory of components anyway. Limiting ourselves to one function per file means many files which are very short, like this one. That means more separate files to open sequentially when attempting to take in the codebase. It also means more redundant license and import statements which, if combined into fewer files, would additionally reduce the overall number of human-relevant lines needed to do a full readthrough of the codebase. My thinking here is partially inspired by the [literate programming perspective](https://github.com/apiad/illiterate/tree/370da18870eefc5d29d579421234ba18296ee829#so-what-does-illiterate-proposes), which can be summarized as "write programs as though the file itself is your entire blog post explaining why the code is needed". It's a lofty goal, but I find the suggestion to think of a codebase as an actual handbook, intended to be read front to back by humans, useful in ensuring code stays navigable.

That all said, combining two files isn't that big of a deal on its own. Do what you feel like!

lahwran (approved):
No bugs or blocking style issues that I see. Provided optional style feedback.

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/reducer.ts
```diff
@@ -0,0 +1,63 @@
+import { ISelectColumnsState, ReplicateSelect } from './types';
```
njbriggs, line 17: This file needs a license

seanfeld: ugh - i gotta look into a linter/ auto thing for us

app/cdap/services/react/customHooks/useDebounce.ts
njbriggs: I'm fine with keeping these as separate. If we keep adding hooks, the file will get large enough to split, and I don't find multiple files in one directory that hard to navigate. Consider also the test files - one set of tests for one hook is about 40 lines (not counting copyright) so that file would grow even quicker than the main file.

lahwran: That's a fair point about relative test size. bikeshed.end() 👍

PR #220, opened 2021-10-18
Adds styled-components

antony-britto:
Adds styled-components


antony-britto:
This is required for subsequent cherry picks which are based on styled-components.

njbriggs (commented):
We don't want to bring this into 6.5 - we want to keep 6.5 to bugfixes only. We will have to duplicate any changes made to styled-components work in the old code base for cherry picks.

antony-britto:
> We don't want to bring this into 6.5 - we want to keep 6.5 to bugfixes only. We will have to duplicate any changes made to styled-components work in the old code base for cherry picks.

This required for another cherry pick - https://github.com/cdapio/cdap-ui/pull/221

njbriggs:
CMEK's not planned for 6.5.1. And if it were, it would be best to not bring a new dependency into a release branch.

PR #227, opened 2021-10-27
podesta - transformations styling

seanfeld:
Basically all of the styling updates for the transformation table

Screen Shot 2021-10-26 at 10 02 12 PM

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SearchBox/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SearchBox/index.tsx
@@ -17,6 +17,8 @@
 import * as React from 'react';
 import withStyles, { WithStyles, StyleRules } from '@material-ui/core/styles/withStyles';
 import TextField from '@material-ui/core/TextField';
+import InputAdornment from '@material-ui/core/InputAdornment';
+import SearchIcon from '@material-ui/icons/Search';
 
 const styles = (): StyleRules => {
   return {
@@ -35,9 +37,16 @@ interface ISearchBoxProps extends WithStyles<typeof styles> {
   value: string;
   placeholder?: string;
   onChange: (value: string) => void;
+  withAdornment?: boolean;
 }
 
-const SearchBoxView: React.FC<ISearchBoxProps> = ({ classes, value, onChange, placeholder }) => {
+const SearchBoxView: React.FC<ISearchBoxProps> = ({
+  classes,
+  value,
+  onChange,
+  placeholder,
+  withAdornment,
+}) => {
   function handleChange(e) {
     onChange(e.target.value);
   }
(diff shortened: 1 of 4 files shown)


app/cdap/components/Replicator/Create/Content/SearchBox/index.tsx
```diff
@@ -49,6 +58,28 @@ const SearchBoxView: React.FC<ISearchBoxProps> = ({ classes, value, onChange, pl
       onChange={handleChange}
       placeholder={placeholder}
       variant="outlined"
+      InputProps={
```
njbriggs, line 61: We should make a note to make this a shared component and use it across the board.

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
```diff
@@ -336,25 +264,34 @@ const SelectColumnsView: React.FC<ISelectColumnsProps> = (props) => {
...
+                />
+                <FormControlLabel
+                  value={ReplicateSelect.individual}
+                  control={<StyledRadio color="primary" />}
+                  label="Select the columns to replicate"
+                />
+              </StyledRadioGroup>
+            </RadioContainer>
+            <RefreshContainer>
+              <Button variant="outlined" color="primary">
```
njbriggs, line 286: We should start creating shared components for our buttons so we can formalize the set of appearance combinations that we use.

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/table.tsx
```diff
@@ -0,0 +1,241 @@
...
+  toggleSelected: (row: any) => void;
+  I18N_PREFIX: string;
+}) => {
+  const numNotSupported = 3;
+  let supportedError;
+  if (numNotSupported > 0) {
+    supportedError = (
+      <>
+        {' - '}
+        <RedSpan>{numNotSupported} not or partially supported</RedSpan>
```
njbriggs: Nit: Don't name the component for the visual properties - could be `EmphasisSpan` or `WarningSpan`.

Could you consider using `Alert`? Also, we don't want color to be the only thing showing this is an issue - could we add an appropriate icon?

njbriggs (commented):
Mostly LGTM - but I think we should start creating shared components.

app/cdap/components/Replicator/Create/Content/SearchBox/index.tsx
seanfeld: agreed with both the points about shared components - i don't know if now is the time to do reorganizing because i still haven't merged that other pr with the components reorganized because i've been concentrating on this

njbriggs (approved):

PR #244, opened 2021-11-12
fix ignoring of click event on navbar items when market modal is active

mrahanjam:
This PR fixes the issue of navigating to other tabs (e.g., Operations, Report, etc.) when the market modal is active.

**Current Behaviour:** Click event is ignored on `navbar` items (except for `HUB` button) when market modal is active.

**Expected Behaviour:** User should be able to exit out of the modal by clicking on other items in the `navbar`.


mrahanjam:
Wrong email address used for cla.

PR #251, opened 2021-11-16
display 0 as the value for scale and precision of decimal fields

mrahanjam:
## **Issue:** [CDAP-18588](https://cdap.atlassian.net/browse/CDAP-18588)

**Current Behaviour:** Setting the scale or precision as 0 for decimal fields in the pipeline studio is captured behind the scene, but not displayed in the UI (fall back to default values).

**Expected Behaviour:** Setting the scale or precision as 0 for decimal fields in the pipeline studio should be reflected in the UI as well.

**Context:** We had a truthy condition check against zero values that would return false, hence falling back to default values.


njbriggs (approved):

flow0ftime:
Hi, we found that this issue still exists in Data Fusion Developer version 6.6.0 with Chrome. Kindly fix this issue asap. Thanks

PR #252, opened 2021-11-16
persist entered values when toggling macro button in pipeline studio …

mrahanjam:
## **Issue:** [CDAP-18385](https://cdap.atlassian.net/browse/CDAP-18385)

This PR ensures previously entered values in pipeline studio configuration fields persist when toggling the macro button

**Current Behaviour:** Toggling the macro button clears the field

**Expected Behaviour:** Toggling the macro button should display previously entered value

### DIFF
--- a/app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
+++ b/app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
@@ -83,6 +83,8 @@ const WIDGET_CATEGORY = 'widget-category';
 
 interface IState {
   isMacroTextbox: boolean;
+  prevValue: string;
+  prevMacroValue: string;
 }
 
 class PropertyRowView extends React.Component<IPropertyRowProps, IState> {
@@ -92,6 +94,8 @@ class PropertyRowView extends React.Component<IPropertyRowProps, IState> {
 
   public state = {
     isMacroTextbox: this.isMacroTextbox(),
+    prevValue: '',
+    prevMacroValue: '${}',
   };
 
   public shouldComponentUpdate(nextProps) {
@@ -122,9 +126,11 @@ class PropertyRowView extends React.Component<IPropertyRowProps, IState> {
     const newValue = !this.state.isMacroTextbox;
 
     if (newValue) {
-      this.handleChange('${}');
+      this.setState({ prevValue: this.props.value });
+      this.handleChange(this.state.prevMacroValue);
     } else {
-      this.handleChange('');
+      this.setState({ prevMacroValue: this.props.value });
+      this.handleChange(this.state.prevValue);
     }
 
     this.setState({ isMacroTextbox: newValue });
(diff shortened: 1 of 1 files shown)


app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
```diff
@@ -83,6 +83,8 @@ const WIDGET_CATEGORY = 'widget-category';
 
 interface IState {
   isMacroTextbox: boolean;
+  prevValue: string;
+  prevMacroValue: string;
```
mrahanjam: Persisting the previously entered macro was not part of the original issue. I figured we could do it since we're persisting non-macro entries. Let me know if it makes sense.

app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
```diff
@@ -92,6 +94,8 @@ class PropertyRowView extends React.Component<IPropertyRowProps, IState> {
 
   public state = {
     isMacroTextbox: this.isMacroTextbox(),
+    prevValue: '',
+    prevMacroValue: '${}',
   };
 
   public shouldComponentUpdate(nextProps) {
```
seanfeld, line 99: oh wow the github collapsed drawer made it look like you were doing this in shouldComponentUpdate haha

app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
```diff
@@ -122,9 +126,11 @@ class PropertyRowView extends React.Component<IPropertyRowProps, IState> {
     const newValue = !this.state.isMacroTextbox;
 
     if (newValue) {
-      this.handleChange('${}');
+      this.setState({ prevValue: this.props.value });
```
seanfeld: shouldn't we only set state once at the end of this? it looks like it'll run setState twice on every toggleMacro?

app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
```diff
@@ -82,7 +82,7 @@ const PLUGIN = 'plugin';
 const WIDGET_CATEGORY = 'widget-category';
 
 interface IState {
-  isMacroTextbox: boolean;
+  [key: string]: string | boolean;
```
mrahanjam, line 85: I had to make this change to keep TS happy about `[valueStateToBeUpdated]` on line 134. I can also keep the `Istate` interface as before and remove the redundant `setState` call like below:

```javascript
    if (newValue) {
      this.setState({ isMacroTextbox: newValue, prevValue: this.props.value });
      this.handleChange(this.state.prevMacroValue);
    } else {
      this.setState({ isMacroTextbox: newValue, prevMacroValue: this.props.value });
      this.handleChange(this.state.prevValue);
    }
```

app/cdap/components/ConfigurationGroup/PropertyRow/index.tsx
mrahanjam: Good catch, missed the `setState` call down below. Addressed in [1887bc9](https://github.com/cdapio/cdap-ui/pull/252/commits/1887bc960b6612b5b692dc8614104bbb50a7b35e)

seanfeld (approved):

PR #254, opened 2021-11-17
fix page nav when setting runtime args in control center

mrahanjam:
## **Issue:** [CDAP-18442](https://cdap.atlassian.net/browse/CDAP-18442)

This PR fixes the page navigation issue when setting runtime arguments for applications in control center

**Current Behaviour:** After setting a runtime arg and clicking on `Save & Close`, control center jumps to the next page

**Expected Behaviour:** After setting a runtime arg and clicking on `Save & Close`, control center should remain on current page

### DIFF
--- a/app/cdap/components/EntityListView/SearchStore/ActionCreator.js
+++ b/app/cdap/components/EntityListView/SearchStore/ActionCreator.js
@@ -28,7 +28,7 @@ import { Theme } from 'services/ThemeHelper';
 
 const search = () => {
   const namespace = getCurrentNamespace();
-  let { offset, limit, activeFilters, activeSort, query, cursor } = SearchStore.getState().search;
+  let { offset, limit, activeFilters, activeSort, query } = SearchStore.getState().search;
 
   let params = {
     namespace: namespace,
@@ -39,7 +39,6 @@ const search = () => {
     query,
     responseFormat: 'v6',
     cursorRequested: true,
-    cursor,
   };
   if (query !== DEFAULT_SEARCH_QUERY) {
     delete params.sort;


app/cdap/components/EntityListView/SearchStore/ActionCreator.js
```diff
@@ -28,7 +28,7 @@ import { Theme } from 'services/ThemeHelper';
 
 const search = () => {
   const namespace = getCurrentNamespace();
-  let { offset, limit, activeFilters, activeSort, query, cursor } = SearchStore.getState().search;
```
mrahanjam, line 31: I'm really not sure about this "fix". I noticed that passing a `cursor` value to the search API results in the response object to contain the `next` set of records. Removing it solves the issue in the referenced bug, but this `search` function is quite popular across modules and I wanna be sure this change is logically correct. (i.e., what does this cursor param do?)

seanfeld: Looks like cursor is just used in the searchStore - did you try logging it or anything to see what is sent as a cursor?

mrahanjam: It's either `undefined` (when navigating to the first page) or has a value like this: 
```
10:10:false::default:application,dataset:creation-time DESC:9223370399852806853:*
```

mrahanjam: I believe it gets populated when you're on the first page (having it `undefined` results in navigating to the first page upon state change). But if you're already on the first page, the cursor has a value, hence an state change (i.e., adding a runtime arg) leads to navigating to the next page.

seanfeld: Hmm that is a super non-obvious value. I guess this is fine though because it seems like its working as intended? Are you trying to merge this pr? Or are you going to try and find a different way or something?

mrahanjam: Yup, I wanted to make sure it's safe to be merged. It does fix the bug and seems to be working okay. I can make sure it passes all the tests and won't break anything before merging it with develop. As for a different way, I couldn't think of any other back when I was working on it, it'd probably have to happen on the backend/api to differentiate between a navigational click and a non-navigational one.

seanfeld (approved):

PR #256, opened 2021-11-22
Adds redirection for metadata summary page & page titles

antony-britto:
1) Adds redirection between Ng & react pages of metadata summary.
2) Adds page titles for metadata pages.

### DIFF
--- a/app/cdap/components/Metadata/SearchSummary/EntityTopBar/index.tsx
+++ b/app/cdap/components/Metadata/SearchSummary/EntityTopBar/index.tsx
@@ -15,6 +15,7 @@
  */
 
 import React, { useState } from 'react';
+import { Redirect } from 'react-router-dom';
 import T from 'i18n-react';
 import styled from 'styled-components';
 import IconButton from '@material-ui/core/IconButton';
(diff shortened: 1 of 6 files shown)


app/cdap/components/Metadata/SearchSummary/EntityTopBar/index.tsx
```diff
@@ -55,16 +57,35 @@ const EntityTab = styled(Tab)`
...
+
+  if (redirectUrl) {
+    return <Redirect to={redirectUrl} />;
+  }
 
-  function handleTabChange(event, newValue: string) {
-    setEntityTypeTab(Number(newValue));
+  function handleTabChange(event, newValue: number) {
+    if (defaultTab !== newValue) {
+      const baseUrl = `/ns/${getCurrentNamespace()}/metadata/${entityType}/${entityId}/`;
```
njbriggs, line 82: Is there any way to leverage the router so we don't need to put URLs in the component?

app/tracker/routes.js
```diff
@@ -145,7 +147,15 @@ angular.module(PKG.name + '.feature.tracker')
             controller: 'TrackerMetadataController',
             controllerAs: 'MetadataController',
             onEnter: function($stateParams) {
-              document.title = `${productName} | Search | ${$stateParams.entityId} | Summary`;
+              // Redirect to react page when the feature is turned on
+              if (window.CaskCommon.ThemeHelper.Theme.isMetadataInReact) {
+                let searchUrl = reactAppUrl.summary.replace('<namespace>', $stateParams.namespace);
```
njbriggs, line 152: Is there any way to use router to avoid manual URL manipulation here?

antony-britto: This is angular app's router. We need to construct the react url and redirect. Couldn't find other way.

njbriggs: Arrrggghhh good point.

njbriggs (approved):

PR #266, opened 2021-12-01
CDAP-19008 Remove websocket usage. Switch React requests to use Rxjs.ajax

njbriggs:
# CDAP-19008 Remove websocket usage. Switch React requests to use Rxjs.ajax

## Description

Switches React REQUESTs to use rxjs.ajax instead of websockets.

To do:
- [x] Verify error handling/corner case handling matches current behavior
- [x] Determine what type of request/response logging we want, or if we can rely on the browser's Network logging
- [x] React POLLing. Will probably use `Subject`s in a manner similar to the current code
- [x] Put behind feature flag
- [ ] ~~Ng1 network interactions take a different path that will need to be updated: https://github.com/cdapio/cdap-ui/blob/develop/app/services/cask-angular-socket-datasource/datasource.js~~ Converting Ng1 will require significant refactoring of existing code and will not be done in this PR.
- [X] Squash commits

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [X] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19008](https://cdap.atlassian.net/browse/CDAP-19008)

## Test Plan
This feature is wired off by default. It can be tested by enabling and using the app (other than Pipeline Studio).

## Screenshots
N/A


app/cdap/services/datasource/DataSourceWebsockets.js
```diff
@@ -0,0 +1,391 @@
+/*
```
de-lan, line 1: Wonder if it's preferred to use ts?

seanfeld: are these files copied from somewhere? maybe a previous pr before they removed websockets?

njbriggs: This one is a copy, but `DataSourceHttp` is new and should be with us for a while. I'll see what it would take to add types.

njbriggs: I've converted `DataSourceHttp` to TS

app/cdap/services/datasource/DataSourceWebsockets.js
```diff
@@ -0,0 +1,391 @@
...
+  request(resource = {}) {
+    const excludeFromHealthCheck = !!resource.excludeFromHealthCheck;
+    let generatedResource = {
+      id: resource.id || uuidV4(),
+      json: resource.json === false ? false : true,
+      method: resource.method || 'GET',
+      suppressErrors: resource.suppressErrors || false,
+    };
+
+    if (resource.body) {
```
seanfeld, line 162: this code is repeated 4? times - could we make a helper?

seanfeld: not all of it though, just .body, .data, .headers

app/cdap/services/datasource/DataSourceHttp.ts
```diff
@@ -0,0 +1,460 @@
...
+  public request(resource: IInboundResource = {}) {
+    const excludeFromHealthCheck = !!resource.excludeFromHealthCheck;
+    const generatedResource: IResource = {
+      id: resource.id || uuidV4(),
+      json: resource.json === false ? false : true,
+      method: resource.method || 'GET',
+      suppressErrors: resource.suppressErrors || false,
+    };
+
+    if (resource.body) {
```
seanfeld, line 204: I was wondering why the next file looks so familiar - disregard that comment about making a helper down there. The websockets file should be removed eventually so its only repeated once

app/cdap/services/datasource/DataSourceWebsockets.js
seanfeld: nvm - this file is a copy and will be deleted :)

app/cdap/services/datasource/DataSourceHttp.ts
```diff
@@ -0,0 +1,460 @@
...
+    const forEachSorted = (obj: IParamsMap, iterator) => {
+      const keys = Object.keys(params).sort();
+      keys.forEach((key) => {
+        iterator.call(null, obj[key], key);
+      });
+      return keys;
+    };
+
+    const encodeUriQuery = (val: string, pctEncodeSpaces?: boolean) => {
+      return encodeURIComponent(val)
```
seanfeld, line 429: is there a better way to do this?

seanfeld (approved):
looks good - i wish we could break datasourcehttp.ts up a bit but there didn't look like any good ways of doing that that were simple.

njbriggs:
> i wish we could break datasourcehttp.ts up a bit but there didn't look like any good ways of doing that that were simple.

Eventually, we can separate polling out. Edwin had an idea for polling to be controlled by the caller of the API, not for each endpoint to be defined as "polling" or "request".

app/cdap/services/datasource/DataSourceHttp.ts
njbriggs: Probably. I filed https://cdap.atlassian.net/browse/CDAP-19018 to track but I won't tackle in this PR.

PR #271, opened 2021-12-06
Pipeline studio: Removes validation for hidden plugin fields

antony-britto:


seanfeld:
was this merged before and then removed?

antony-britto:
> was this merged before and then removed?

Yes. Reverted and made a fix.

seanfeld (approved):

PR #276, opened 2021-12-10
Connection creation: Fixes tab content re-rendering

antony-britto:


antony-britto:
Since we render only the active tab, users loose their information (input text that they have entered etc...) while toggling between tabs. Hence rendering all tabs and this can be enabled selectively using 'renderAllTabs' props.

mrahanjam (approved):
LGTM!

PR #279, opened 2021-12-14
Adds min & max range for pipeline preview records

antony-britto:


njbriggs:
Is there a Jira for this? Can you add a screenshot?

antony-britto:
image

antony-britto:
Jira link - https://cdap.atlassian.net/browse/CDAP-18668

njbriggs:
I'm wary of using a slider only because it's not as accessible. Can we add a text input? Or just keep the current text input with a range check?

Also, what happens to pipelines that have a value great than 5000? I don't think there are many, but we should know what will happen.

antony-britto:
> I'm wary of using a slider only because it's not as accessible. Can we add a text input? Or just keep the current text input with a range check? 
   > Restored it to existing text input.

> Also, what happens to pipelines that have a value great than 5000? I don't think there are many, but we should know what will happen. 
  > Max limit will be provided by API. UI has a fallback value as 5000 and it will be overridden through an API call.

seanfeld (commented):
Code looks fine but does this address @njbriggs 's concerns about accessibility/usability?

antony-britto:
> Code looks fine but does this address @njbriggs 's concerns about accessibility/usability?

i have reverted to the existing text input. Not using range slider anymore.

image

seanfeld (approved):

PR #287, opened 2021-12-23
BUILD ERROR - explicitly install npmcli/fs globally before node gyp

seanfeld:
# Fix npmcli/fs build error

## Description
explicitily install npmcli/fs@1.0.0 globally before node gyp so node gyp doesn't install npmcli/fs@1.1.0 - need to switch to github actions or this will keep happening

## PR Type
- [x] Bug Fix
- [ ] Feature
- [x] Build Fix
- [ ] Testing
- [ ] General Improvement

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -223,6 +223,15 @@
                 </configuration>
               </execution>
               <execution>
+                <id>install-npmcli-fs</id>
+                <goals>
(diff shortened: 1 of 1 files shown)


sechegaray (approved):
It should be a build fix correct? (In the comment)

PR #289, opened 2021-12-27
Fix z-index of BQ pushdown connection popup

antony-britto:
# TITLE

## Description
Fixing z-index of parent modal to make the nested ones visibles.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement

## Links
Jira: [1020](https://cdap.atlassian.net/browse/PLUGIN-1020)

## Screenshots
Issue:
image

### DIFF
--- a/app/cdap/components/PipelineConfigurations/PipelineConfigurations.scss
+++ b/app/cdap/components/PipelineConfigurations/PipelineConfigurations.scss
@@ -22,6 +22,10 @@ $top-panel-height: 54px;
 $right-position: -360px;
 $sidepanel-width: 170px;
 
+.pipeline-config-modal {
+  z-index: 1060; // overriding default 1300 to make nested modals (1061) visible.
+}
+
 .pipeline-configurations-content {
   width: $modeless-width;
   height: $modeless-height;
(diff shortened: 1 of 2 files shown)


app/cdap/components/PipelineConfigurations/PipelineConfigurations.scss
```diff
@@ -22,6 +22,10 @@ $top-panel-height: 54px;
 $right-position: -360px;
 $sidepanel-width: 170px;
 
+.pipeline-config-modal {
+  z-index: 1060; // overriding default 1300 to make nested modals (1061) visible.
```
seanfeld, line 26: we need to figure out a better long term solution from this in the future @njbriggs

seanfeld (approved):

app/cdap/components/PipelineConfigurations/PipelineConfigurations.scss
antony-britto: One solution could be having a JS counter variable and increment/decrement as we open/hide modals.

PR #303, opened 2022-01-10
Restore default values while toggling connection field in BQ pushdown…

antony-britto:
Issue:
    While toggling connections field we restore default values which were fetched as part of filter configuration groups(FCG). But for deployed pipelines FCG doesn't hold any default values.

Fix:
     Passing default values explicitly instead of relying on filter configuration groups.

## Description
Restore default values while toggling connection field in BQ pushdown config modal.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement

## Links
Jira: [1026](https://cdap.atlassian.net/browse/PLUGIN-1026)

## Test Plan
Added unit tests

## Screenshots
issue:
image

fix:
image

### DIFF
--- a/app/cdap/components/shared/ConfigurationGroup/utilities/index.ts
+++ b/app/cdap/components/shared/ConfigurationGroup/utilities/index.ts
@@ -382,17 +382,18 @@ export function countErrors(
  *
  * @param values latest field values
  * @param filteredConfigurationGroups configuration-groups from Widget JSON
- * @param addDefaults add default values for visible fields that are empty
+ * @param defaultValues default field values
  */
-export function removeFilteredProperties(values, filteredConfigurationGroups, addDefaults = false) {
+export function removeFilteredProperties(values, filteredConfigurationGroups, defaultValues = {}) {
   const newValues = { ...values };
   if (filteredConfigurationGroups) {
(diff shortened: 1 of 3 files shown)


app/cdap/components/shared/ConfigurationGroup/utilities/index.ts
```diff
@@ -382,17 +382,18 @@ export function countErrors(
  *
  * @param values latest field values
  * @param filteredConfigurationGroups configuration-groups from Widget JSON
- * @param addDefaults add default values for visible fields that are empty
+ * @param defaultValues default field values
  */
-export function removeFilteredProperties(values, filteredConfigurationGroups, addDefaults = false) {
+export function removeFilteredProperties(values, filteredConfigurationGroups, defaultValues = {}) {
```
njbriggs, line 387: Nit: Consider `baseValues` instead of `defaultValues`? Since they're not always the "defaults" as they were before.

njbriggs (approved):
LGTM. One possible rename nit but if you don't think it makes the code more clear, then don't worry about it

seanfeld (approved):

PR #321, opened 2022-01-26
Feature flags using react context api

seanfeld:
# React Context Feature Flags

## Description
Adds feature flags using the react context api and adds the feature flag support for replication transformations. This also shows how to use the feature flag hook from inside a class component using the higher order component pattern.

Unfortunately CDAP UI currently gets all configuration from the window in `server/express.js` so making use of that to have these feature flags from the backend.

I also removed `<If></If>` from main.js while working on other things. 

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
@@ -32,7 +32,7 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Heading, { HeadingTypes } from 'components/shared/Heading';
 import ManualSelectTable from 'components/Replicator/Create/Content/SelectTables/ManualSelectTable';
 import SearchBox from 'components/Replicator/Create/Content/SearchBox';
-import { useFeatureFlag } from 'services/react/customHooks/useFeatureFlag';
+import { useFeatureFlagDefaultFalse } from 'services/react/customHooks/useFeatureFlag';
 import debounce from 'lodash/debounce';
 import classnames from 'classnames';
 import {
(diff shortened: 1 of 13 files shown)


mrahanjam (approved):
LGTM 🚀

app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
```diff
@@ -706,4 +706,9 @@ class SelectTablesView extends React.PureComponent<ISelectTablesProps, ISelectTa
 
 const StyledSelectTables = withStyles(styles)(SelectTablesView);
 const SelectTables = createContextConnect(StyledSelectTables);
-export default SelectTables;
+
+// Higher Order Component wrapping class component so we can use useFeatureFlag hook
+export default ({ children, ...props }) => {
+  props.useReplicationTransformation = useFeatureFlag('replication.transformations.enabled');
+  return <SelectTables {...props} />;
```
njbriggs: Nit: Should we make this
```
return <SelectTables
  {{
    ...props,
    useReplicationTransformation: useFeatureFlag('replication.transformations.enabled')
  }}
/>
```
?

I know it's short HOC but mutating `props` is still bugging me.

app/cdap/services/react/customHooks/useFeatureFlag.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Returns true if feature if enabled, false if the flag
+ * does not exist or disabled
+ * Must wrap parent component with services/react/providers/featureFlagProvider
+ * @param name feature flag name
+ * @returns boolean
+ */
+
+export function useFeatureFlag(name: keyof IFeatureFlags | string): boolean {
```
njbriggs: Maybe late to be asking this, but: Was there consideration for forcing disable of features? That is, this design represents features that are by default OFF and can be forced ON with a flag. Did we think about features that are by default ON and can be forced OFF with a flag? Or just create that behavior with a flag called `disableFoo` and an inverted test?

app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
seanfeld: Yeah absolutely I like that a lot better.

njbriggs (approved):

PR #351, opened 2022-02-10
JIRA 18844 - adds error handling for the details page in replication

seanfeld:
# Fix 18844 - handle errors on replication details page in replication

## Description
actually handle errors instead of just console log them

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [18844](https://cdap.atlassian.net/browse/CDAP-18844)

### DIFF
--- a/app/cdap/components/Replicator/Detail/index.tsx
+++ b/app/cdap/components/Replicator/Detail/index.tsx
@@ -32,6 +32,7 @@ import LoadingSVGCentered from 'components/shared/LoadingSVGCentered';
 import ContentHeading from 'components/Replicator/Detail/ContentHeading';
 import { ITableInfo, IArtifactInfo } from '../types';
 import DownloadFile from 'services/download-file';
+import ErrorBanner from 'components/shared/ErrorBanner';
 
 export const DetailContext = React.createContext<Partial<IDetailState>>({});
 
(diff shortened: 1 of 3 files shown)


app/cdap/components/Replicator/Detail/index.tsx
```diff
@@ -104,6 +105,7 @@ interface IDetailState {
   setTimeRange: (timeRange: string) => void;
   exportPipeline: () => void;
   duplicatePipeline: () => void;
+  error: object | string | null;
```
mrahanjam, line 108: Not sure about the error object props, but shouldn't we send `error.response` or `error.message` (whichever is in the error object) and remove `object` type here (i.e., restrict it to string | null) for better readability.

seanfeld: error banner handles that

seanfeld:
```
  const errorMessage: string =
    objectQuery(error, 'response', 'message') || objectQuery(error, 'response') || error;
    ```

mrahanjam: Ah I see 👍

mrahanjam (approved):
LGTM!

app/cdap/components/Replicator/Detail/index.tsx
seanfeld: If we wanted to do this right, the API handler should handlers should all return the same type of error and we shouldn't have to worry about it

PR #364, opened 2022-02-16
CDAP-18675 Source parsing in Wrangler

njbriggs:
# CDAP-18675 Source parsing in Wrangler

## Description
Adds source parsing to Wrangler. Source parsing means the source plugin parses the file before it goes to Wrangler, so the user doesn't need to apply `parse-as-csv` or similar directives.
* The dialog will be shown as determined by the connection's response to browse requests. GCS connection currently provides this. Some sources may get this later (e.g. File) and some will never get it (sources that connect to data that is already structured, like BigQuery and databases).
* The selected configuration will be added to the workspace, which will then be passed through to the source plugin in the pipeline.

TODO
- [X] Look at source browse behavior - don't show dialog
- [X] Look at sink behavior - no change needed
- [X] Cypress test updates
- [ ] File encoding not being set correctly - back end fix; shouldn't block this
- [X] Squash commits

## PR Type
- [ ] Bug Fix
- [X] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18675](https://cdap.atlassian.net/browse/CDAP-18675)

## Test Plan
TBD. Will need attention during bug bash

## Screenshots

Config modal:
image

Loading:
image

### DIFF
--- a/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
+++ b/app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
@@ -36,6 +36,8 @@ import EmptyMessageContainer from 'components/EmptyMessageContainer';
 import ErrorBanner from 'components/shared/ErrorBanner';
 import { getApiErrorMessage } from './apiHelpers';
 import { getConnectionPath } from 'components/Connections/helper';
+import ParsingConfigModal from 'components/Connections/Browser/ParsingConfigModal';
+import keyBy from 'lodash/keyBy';
 
 const PREFIX = 'features.DataPrep.DataPrepBrowser.GenericBrowser';
 import { Redirect } from 'react-router';
@@ -74,6 +76,7 @@ export function GenericBrowser({ initialConnectionId, onEntityChange, selectedPa
   const [currentConnection, setCurrentConnection] = useState(initialConnectionId);
   const [entities, setEntities] = useState([]);
   const [propertyHeaders, setPropertyHeaders] = useState([]);
+  const [sampleProperties, setSampleProperties] = useState({});
   const [totalCount, setTotalCount] = useState(0);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState(null);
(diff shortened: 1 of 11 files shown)


app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
```diff
@@ -74,13 +76,16 @@ export function GenericBrowser({ initialConnectionId, onEntityChange, selectedPa
   const [currentConnection, setCurrentConnection] = useState(initialConnectionId);
   const [entities, setEntities] = useState([]);
   const [propertyHeaders, setPropertyHeaders] = useState([]);
+  const [sampleProperties, setSampleProperties] = useState({});
```
seanfeld, line 79: yeesh this file has so many use states - should convert these to a use reducer eventually (not asking for action)

app/cdap/components/Connections/Browser/ParsingConfigModal/store.tsx
```diff
@@ -0,0 +1,222 @@
...
+    connectionId: connection,
+  };
+
+  const body = {
+    path: entity.path,
+    properties: {},
+  };
+
+  let widgetKey;
+  const widgetObs = ConnectionsApi.getSpecification(params, body).pipe(
```
seanfeld, line 142: can we get some context comments

app/cdap/components/Connections/Browser/ParsingConfigModal/store.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+            // set default value
+            if (!selectedProperties[propertyWidget.name] && propertyWidget['widget-attributes']) {
+              selectedProperties[propertyWidget.name] = propertyWidget['widget-attributes'].default;
+            }
+          }
+        });
+      });
+
+      dispatch({
```
seanfeld, line 192: can we combine all these into one dispatch call? `SET_INITIAL_PROPERTIES` or something like that and set all the values at once

seanfeld (approved):
just nits

app/cdap/components/Connections/Browser/GenericBrowser/index.tsx
njbriggs: Acknowledged

mrahanjam (approved):
LGTM!

PR #374, opened 2022-02-23
fix: add a clearer error message when database is not found

de-lan:
# Issue when you create a replication job with incorrect database name

## Description
it’s actually more a bad UI pattern rather than a bug. When database is not found in the replication assessment, the UI needs users to manually put in the correct database name and table(s) name, once both fields are filled, users can proceed. We added a clearer message to mitigate the obscurity. The best solution should be assessing the draft right after the "Configure source" tab. Will be in a new ticket.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18781](https://cdap.atlassian.net/browse/CDAP-18781)

## Screenshots
image

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
@@ -26,7 +26,7 @@ import orderBy from 'lodash/orderBy';
 import LoadingSVGCentered from 'components/shared/LoadingSVGCentered';
 import SelectColumns from 'components/Replicator/Create/Content/SelectColumns';
 import SelectColumnsWithTransforms from 'components/Replicator/Create/Content/SelectColumnsWithTransforms';
-import { extractErrorMessage } from 'services/helpers';
+import { extractErrorMessage, isUnknownDatabaseError } from 'services/helpers';
 import { generateTableKey, getTableDisplayName } from 'components/Replicator/utilities';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Heading, { HeadingTypes } from 'components/shared/Heading';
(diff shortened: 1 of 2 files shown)


app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
```diff
@@ -450,6 +450,12 @@ class SelectTablesView extends React.PureComponent<ISelectTablesProps, ISelectTa
         <div className="text-danger">
           <Heading type={HeadingTypes.h5} label="Error" />
           <span>{this.state.error}</span>
+          <div>
```
seanfeld: is this the only type of error that can be returned? I think you need to check if the database name is incorrect and conditionally add more error message

seanfeld (changes requested):

app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
de-lan: You're right, will update accordingly

app/cdap/services/helpers.js
```diff
@@ -506,6 +506,10 @@ function extractErrorMessage(errObj) {
   return objectQuery(errObj, 'response', 'message') || objectQuery(errObj, 'response');
 }
 
+function isUnknownDatabaseError(err) {
```
seanfeld, line 514: for new exported functions we do jsdoc style comments like https://github.com/cdapio/cdap-ui/blob/62c7fbe190e5f0055454feb94e85c021f35dcaad/app/cdap/services/react/customHooks/useFeatureFlag.ts#L34

seanfeld (approved):

PR #376, opened 2022-02-23
[CDAP-17955] Do not block draft deployment for warnings in assessment

de-lan:
# Do not block draft deployment for warnings in assessment

## Description
Currently, if users have some database or table name that doesn't meet the requirement of Big Query, users will get a warning in assessment saying that those names will be normalized. However for such warning, we should not block the deployment of the draft.

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-17955](https://cdap.atlassian.net/browse/CDAP-17955)

## Screenshots
image

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/Assessment/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/Assessment/index.tsx
@@ -14,13 +14,12 @@
  * the License.
  */
 
-import * as React from 'react';
+import React, { useState } from 'react';
 import withStyles, { WithStyles, StyleRules } from '@material-ui/core/styles/withStyles';
 import { createContextConnect, ICreateContext } from 'components/Replicator/Create';
 import { getCurrentNamespace } from 'services/NamespaceStore';
 import { MyReplicatorApi } from 'api/replicator';
 import TablesAssessment from 'components/Replicator/Create/Content/Assessment/TablesAssessment';
-import If from 'components/shared/If';
 import classnames from 'classnames';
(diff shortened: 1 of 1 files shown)


seanzhougoogle (approved):

app/cdap/components/Replicator/Create/Content/Assessment/index.tsx
```diff
@@ -97,6 +102,7 @@ const AssessmentView: React.FC<ICreateContext & WithStyles<typeof styles>> = ({
   const [error, setError] = React.useState(null);
   const [schemaErrorCount, setSchemaErrorCount] = React.useState(0);
   const [view, setView] = React.useState(VIEWS.tables);
+  const [errorFeaturesCount, setErrorFeaturesCount] = React.useState(0);
```
seanfeld: Import react, {useState} from 'react'; - i know it was previously done with React.useState but any time you alter the code please update the rest of the useState calls.

Also if you alter code with an <If></If> component, please remove it that replace it with { condition && <>}.

Last thing - if you have more than 4 useState calls, please replace it with a useReducer (don't do it on this component) for new components.

de-lan: Thanks! Updated

seanfeld (approved):
Thanks for updating the code! if we keep doing stuff like this we reduce our tech debt :)

PR #382, opened 2022-03-03
CDAP-18170: Add manual check for MySQL JDBC

sumengwang:
# CDAP-18170: Add manual check for MySQL JDBC

## Description
Add a manual check to see if user has input JDBC plugin for MySQL connection when clicking test connection, preventing unnecessary api call

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18170](https://cdap.atlassian.net/browse/CDAP-18170)

## Test Plan
manual test

### DIFF
--- a/app/cdap/components/Connections/Create/index.tsx
+++ b/app/cdap/components/Connections/Create/index.tsx
@@ -204,25 +204,30 @@ export function CreateConnection({
     setConfigurationErrors(null);
     setTestInProgress(true);
 
-    try {
-      const testResult = await testConnection(connectionConfiguration);
-      setTestResponseMessages(testResult);
-      setTestSucceeded(!testResult);
-
-      if (testResult) {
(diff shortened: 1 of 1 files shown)


njbriggs (commented):
I don't think this is the right fix, as we don't want to hard-code behavior for a specific connection type in the UI. Let's talk on Monday

njbriggs:
We should be able to check the required fields based on the data available in the component. `connectionDetails.connectorProperties` should hold information about which fields are required. Then you can construct an error object to highlight fields with missing values (see how the error object is structured when create fails).

sumengwang:
This required field is false for MySQL's JDBC plugin, does it mean I need to modify the api that returns this property?
image

PR #384, opened 2022-03-05
CDAP-17954: add dynamic hiding overflow values for multiSelect

sumengwang:
# CDAP-17954: add dynamic hiding overflow values for multiSelect

## Description
Changed the logic of how MultiSelect renders values. Before it shows <firstElement> + <totalcount - 1>. Now it will try to fill out the available space then show the remaining count number (see attached screenshots)

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-17954](https://cdap.atlassian.net/browse/CDAP-17954)

## Test Plan
Manual test

## Screenshots
image
image

### DIFF
--- a/app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
+++ b/app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
@@ -20,6 +20,8 @@ import Checkbox from '@material-ui/core/Checkbox';
 import { IWidgetProps } from 'components/AbstractWidget';
 import ListItemText from '@material-ui/core/ListItemText';
 import MenuItem from '@material-ui/core/MenuItem';
+import Box from '@material-ui/core/Box';
+import Chip from '@material-ui/core/Chip';
 import Select from '@material-ui/core/Select';
 import { WIDGET_PROPTYPES } from 'components/AbstractWidget/constants';
 import { objectQuery } from 'services/helpers';
@@ -41,6 +43,9 @@ const styles = (theme) => {
   return {
     root: {
       margin: theme.Spacing(2),
+    },
+    chip: {
+      margin: 1,
     },
   };
 };
@@ -74,7 +79,7 @@ function MultiSelectBase({
 
   // Get the width of the select box for different window size
   const ref = useRef(null);
-  const [selectWidth, setSelectWidth] = useState(500);
+  const [selectWidth, setSelectWidth] = useState(600);
 
   //  onChangeHandler takes array, turns it into string w/delimiter, and calls onChange on the string
   const onChangeHandler = (event: React.ChangeEvent<HTMLSelectElement>) => {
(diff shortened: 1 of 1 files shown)


bdmogal:
@sumengwang tiny feedback: Is there anyway to make the count not look like one of the options? Also, instead of ``+10``, can we say ``10 more``, and have it be a link that opens a tooltip with the other selections?

sumengwang:
@bdmogal I'll incorporate your feedbacks. Thanks

sumengwang:
I changed the selections style from text to cards, and differentiate "...n more" with different color, so that users won't think it's a header.
image

sumengwang:
changed 'n more' to a blue hyperlink style, when hovering on it, it will show a tooltip containing the hidden selections. 

image

app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
```diff
@@ -42,9 +39,26 @@ const styles = (theme) => {
...
+      backgroundColor: 'white',
+    },
+    hyperlink: {
+      margin: 2,
+      color: '#0000EE',
+    },
   };
 };
 
+const LightTooltip = withStyles((theme) => ({
```
seanfeld: Can you do this with a styled component instead?

app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
```diff
@@ -90,23 +120,40 @@ function MultiSelectBase({
...
     }
-    return `${selectionLabel} ${additionalSelectionCount}`;
+    return (
+      <Box>
+        {shownSelections.map((value) => (
+          <Chip className={classes.chip} variant="outlined" key={value} label={value} />
+        ))}
+        {additionalSelectionCount != '' && (
+          <LightTooltip title={additionalSelectionText} placement="right-start">
+            <u className={classes.hyperlink}>{additionalSelectionCount}</u>
```
seanfeld: I'm not sure this is the correct usage for <u> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u - is this just to apply an underline styling? If so this should be a <span>.

seanfeld (changes requested):
Check the comment about <u> and then for _new_ styling we should be using styled components.

app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
sumengwang: Yes the u tag is just to provide a underline

app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
sumengwang: Having trouble to use styled component to override Tooltip styles

mrahanjam: What if you use something like:
```javascript
export const StyledTooltip = styled(Tooltip)`
    background-color: ${(props) => props.theme.palette.white},
    color: rgba(0, 0, 0, 0.87),
    box-shadow: ${(props) => props.theme.shadows[1]},
    font-size: 11,
`
```

seanfeld: i think he was saying he couldn't get any of the styles to render with tooltip - i think he needs to do something like this
```
const GroupsContainer = styled.div`
  && {
    && {
      && {
        height: auto;
        overflow: scroll;
        position: absolute;
        right: 1px;
      }
    }
  }
`;
```

seanfeld: ^ specificity issue

sumengwang: Yea, I was only able to override the Tooltip styles with 'withStyles'

app/cdap/components/AbstractWidget/FormInputs/MultiSelect/index.tsx
```diff
@@ -42,6 +38,20 @@ const styles = (theme) => {
     root: {
       margin: theme.Spacing(2),
     },
+    chip: {
```
seanfeld: Okay sorry if I wasn't clear, none of these things should be done this way. We decided _new_ stuff uses styled components and all these are new. If you can't get something to work, like the tooltip, its fine to do that the other way but you should add a comment mentioning that. Chip and hyperlink should be done using styled components

sumengwang: made the changes

seanfeld (approved):

PR #400, opened 2022-04-01
[CDAP-18971 | CDAP-19054] Disable mask option when data type is not VARCHAR

sumengwang:
# [CDAP-18971 | CDAP-19054] Disable musk option when data type is not VARCHAR

## Description
disable non string column's mask option
fix target data type not showing correctly

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18971](https://cdap.atlassian.net/browse/CDAP-18971) [CDAP-19054](https://cdap.atlassian.net/browse/CDAP-19054)

## Screenshots
image
image

[CDAP-18971]: https://cdap.atlassian.net/browse/CDAP-18971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
@@ -39,6 +39,8 @@ export default function TransformAddButton({
   const open = !!anchorEl;
   const subMenuOpen = !!subMenuAnchorEl;
 
+  const isString = row.targetType.toLowerCase() === 'string';
+
   const handleClick = (event) => {
     setAnchorEl(event.currentTarget);
   };
@@ -135,7 +137,7 @@ export default function TransformAddButton({
         <MenuItem onClick={handleMenuClick}>
           Rename <ArrowRight />
         </MenuItem>
-        <MenuItem onClick={handleMaskOpen}>
+        <MenuItem disabled={!isString} onClick={handleMaskOpen}>
           Mask <ArrowRight />
         </MenuItem>
         <Menu
@@ -154,9 +156,13 @@ export default function TransformAddButton({
             dense: true,
           }}
         >
-          <MenuItem onClick={() => handleSetMaskLast(2)}>Show last 2</MenuItem>
-          <MenuItem onClick={() => handleSetMaskLast(4)}>Show last 4</MenuItem>
-          <MenuItem onClick={handleMenuClick}>
+          <MenuItem disabled={!isString} onClick={() => handleSetMaskLast(2)}>
+            Show last 2
+          </MenuItem>
+          <MenuItem disabled={!isString} onClick={() => handleSetMaskLast(4)}>
+            Show last 4
+          </MenuItem>
+          <MenuItem disabled={!isString} onClick={handleMenuClick}>
             Custom <ArrowRight />
           </MenuItem>
           {tinkEnabled && (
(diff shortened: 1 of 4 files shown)


app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
```diff
@@ -135,9 +135,16 @@ export default function TransformAddButton({
...
-        <MenuItem onClick={handleMaskOpen}>
-          Mask <ArrowRight />
-        </MenuItem>
+        {row.type.toLowerCase() === 'varchar' && (
+          <MenuItem onClick={handleMaskOpen}>
+            Mask <ArrowRight />
+          </MenuItem>
+        )}
+        {row.type.toLowerCase() !== 'varchar' && (
+          <MenuItem disabled onClick={handleMaskOpen}>
```
seanfeld: this should be to disabled={row.type.toLowerCase() !== 'varchar'} instead of having two different ones

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
```diff
@@ -154,11 +161,28 @@ export default function TransformAddButton({
             dense: true,
           }}
         >
-          <MenuItem onClick={() => handleSetMaskLast(2)}>Show last 2</MenuItem>
-          <MenuItem onClick={() => handleSetMaskLast(4)}>Show last 4</MenuItem>
-          <MenuItem onClick={handleMenuClick}>
-            Custom <ArrowRight />
-          </MenuItem>
+          {row.type.toLowerCase() === 'varchar' && (
```
seanfeld: Same thing here - disabled={row.type.toLowerCase !== 'varchar'}

seanfeld (changes requested):
are we sure that varchar is the only type? like are there no char, string etc?

sumengwang:
> are we sure that varchar is the only type? like are there no char, string etc?

I added an exhaust list of string types. Let me know your thoughts.

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
```diff
@@ -135,7 +135,10 @@ export default function TransformAddButton({
         <MenuItem onClick={handleMenuClick}>
           Rename <ArrowRight />
         </MenuItem>
-        <MenuItem onClick={handleMaskOpen}>
+        <MenuItem
+          disabled={!javaSqlStringTypes.includes(row.type.toLowerCase())}
```
de-lan: Could we extract out this condition check as a reusable function instead of repeat the code four times?

seanzhougoogle:
> > are we sure that varchar is the only type? like are there no char, string etc?
> 
> I added an exhaust list of string types. Let me know your thoughts.

checking source type may not be correct, because source type is depending on source plugin and JDBC implementation. And the list seems not correct, e.g. why `blob` is in the list ?

although checking target is not robust, either, but for now we have only one target

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
```diff
@@ -112,29 +113,36 @@ const SelectColumnsView = (props: ISelectColumnsProps) => {
       body.schema = props.tableInfo.schema;
     }
 
-    MyReplicatorApi.getTableInfo(params, body).subscribe(
-      (res) => {
-        const selectedColumns = getInitialSelectedColumns(res.columns);
-
+    Observable.forkJoin([
```
de-lan, line 116: 👍

de-lan (approved):
LGTM

seanfeld (approved):

PR #411, opened 2022-04-08
[CDAP-18958] Make replicator transformation steps remember the column rename

sumengwang:
# [CDAP-18958] Make replicator transformation steps remember the column rename

## Description
Before if a user does a chain of rename operations to the same column, the directive will only show the original 'from', not the intermediate values. Now it will check the transformation history and find out the current column name if a rename ever happened previously.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18958](https://cdap.atlassian.net/browse/CDAP-18958)

## Test Plan
manual testing

## Screenshots
image

[CDAP-18958]: https://cdap.atlassian.net/browse/CDAP-18958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
@@ -27,7 +27,7 @@ export default function TransformAddButton({
   row,
   addColumnsToTransforms,
   tinkEnabled,
-  transforms,
+  currentColumnName,
 }: ITransformAddProps) {
   // todo replace this with a useReducer
   const [anchorEl, setAnchorEl] = useState(null);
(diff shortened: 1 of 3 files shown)


app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/TransformAdd.tsx
```diff
@@ -39,6 +40,17 @@ export default function TransformAddButton({
   const open = !!anchorEl;
   const subMenuOpen = !!subMenuAnchorEl;
 
+  const getCurrentColumnName = (columnName: string) => {
```
seanfeld: can we move this out of here and just pass the current column name to TransformAdd - that way you don't have to pass all of the transformations to that component

seanfeld (changes requested):

seanfeld (approved):

PR #421, opened 2022-04-12
[CDAP-18581] improve error checking for import connection json

sumengwang:
# [CDAP-18581] improve error checking for import connection json

## Description
- Add error checking for importing connection from a json file. QUESTION: as of now if there's an error, it will still open the createConnectionModal, do we want this or we want to close the modal altogether?
- Allow importing the same file consecutively, before the onChange hook will not fire if same file is uploaded

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18581](https://cdap.atlassian.net/browse/CDAP-18581)

## Screenshots
image

[CDAP-18581]: https://cdap.atlassian.net/browse/CDAP-18581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/Connections/Create/index.tsx
+++ b/app/cdap/components/Connections/Create/index.tsx
@@ -27,11 +27,11 @@ import {
   navigateToConfigStep,
   navigateToConnectionCategoryStep,
   navigateToConnectionList,
-  IConnectorDetails,
   fetchConnectionDetails,
   createConnection,
   getConnection,
   testConnection,
+  IConnectorDetailsWithErrorMessage,
 } from 'components/Connections/Create/reducer';
 import LoadingSVGCentered from 'components/shared/LoadingSVGCentered';
 import { Redirect } from 'react-router';
@@ -69,10 +69,13 @@ export function CreateConnection({
   const [loading, setLoading] = React.useState(true);
   const [state, dispatch] = React.useReducer(reducer, initialState);
   const activeCategory = React.useRef(null);
-  const [connectionDetails, setConnectionDetails] = React.useState<IConnectorDetails>({
+  const [connectionDetails, setConnectionDetails] = React.useState<
+    IConnectorDetailsWithErrorMessage
+  >({
     connectorWidgetJSON: null,
     connectorProperties: null,
     connectorDoc: null,
+    error: null,
   });
   const [initValues, setInitValues] = React.useState({});
   const [error, setError] = React.useState(null);
@@ -127,11 +130,13 @@ export function CreateConnection({
     setLoading(true);
     const connDetails = await fetchConnectionDetails(selectedConnector);
     setConnectionDetails(connDetails);
-    navigateToConfigStep(dispatch, selectedConnector);
+    if (!connDetails.error) {
+      navigateToConfigStep(dispatch, selectedConnector);
+    }
     setLoading(false);
     setTestResponseMessages(undefined);
     setTestSucceeded(false);
-    setError(null);
+    setError(connDetails.error);
     setConfigurationErrors(null);
   };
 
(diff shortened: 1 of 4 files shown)


app/cdap/components/Connections/Create/reducer.ts
```diff
@@ -46,6 +46,10 @@ export interface IConnectorDetails {
   connectorDoc: unknown;
 }
 
+export interface IConnectorDetailsWithErrorMessage extends IConnectorDetails {
```
sumengwang: I was not sure whether I should just add the new field into the original IConnectorDetails

app/cdap/components/Connections/ImportConnectionBtn/index.tsx
```diff
@@ -38,6 +38,12 @@ export default function ImportConnectionBtn({ onCreate, className = null }) {
 
   const classes = useStyle();
 
+  function handleFileClear() {
```
sumengwang, line 42: This allows onChange to fire for same file

app/cdap/components/Connections/ImportConnectionBtn/index.tsx
```diff
@@ -85,6 +91,7 @@ export default function ImportConnectionBtn({ onCreate, className = null }) {
         <input
           type="file"
           accept=".json"
+          onClick={handleFileClear}
           onChange={handleFile}
```
de-lan, line 96: LGTM
One nitpick is that when we meet `<If>` tag, we could always replace them with {CONDITION_STATEMENT && <>}, just to kill some of the tech debt

de-lan, line 96: LGTM
One nitpick is that when we meet `<If>` tag, we could always replace them with {CONDITION_STATEMENT && <>}, just to kill some of the tech debt

de-lan (approved):

app/cdap/components/Connections/Create/index.tsx
```diff
@@ -69,10 +69,13 @@ export function CreateConnection({
   const [loading, setLoading] = React.useState(true);
   const [state, dispatch] = React.useReducer(reducer, initialState);
   const activeCategory = React.useRef(null);
-  const [connectionDetails, setConnectionDetails] = React.useState<IConnectorDetails>({
+  const [connectionDetails, setConnectionDetails] = React.useState<
```
seanfeld: If you modify `React.useState` please replace it with `useState` and `import {useState} from react` in the future.

app/cdap/components/Connections/Create/reducer.ts
seanfeld: Is this used in all IconnectorDetails? if so feel free to add it? Unsure why you're not sure

app/cdap/components/Connections/ImportConnectionBtn/index.tsx
seanfeld: why not add this as a comment to the code?

seanfeld (approved):
Couple of nits

app/cdap/components/Connections/Create/index.tsx
sumengwang: Is this tech debt? Should I just modify all while I'm at it?

seanfeld: It definitely is tech debt - i usually modify any instances of this when i see it and am modifying that file because theres no way to screw it up and its very easy to do. 

Basically i look at it like there are two types of code tech debt: 
- Small pieces of tech debt that should be changed any time you see them _and_ modify the file like `react.useState` and `<If />`
- Larger pieces like converting from class components or react.createClass components to functional + hooks should only be done if you're heavily modifying the file (use your judgement here).

PR #447, opened 2022-04-20
[CDAP-18952] fix mssql inconsistent table name

sumengwang:
# [CDAP-18952] fix mssql inconsistent table name

## Description
After changing the format of mssql table name to `schema.table`, the response table name will also become this format, but the UI only need the tablename to do subsequent work. Hence need the extra split step

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18952](https://cdap.atlassian.net/browse/CDAP-18952)

[CDAP-18952]: https://cdap.atlassian.net/browse/CDAP-18952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/Replicator/utilities/index.ts
+++ b/app/cdap/components/Replicator/utilities/index.ts
@@ -446,7 +446,12 @@ export function identifyReplicatorEntityFromMetadata(metadata: {
  */
 export function parseErrorMessageForTransformations(message: string) {
   const [_, tablePiece, columnPiece, messagePiece] = message.split(' : ');
-  const table = tablePiece.split(' ')[0];
+  // for mssql source, table name will be in format of schema.table
+  // hence need to split on '.' and take the last element as table name
+  const table = tablePiece
+    .split(' ')[0]
+    .split('.')
+    .slice(-1)[0];
   const column = columnPiece.split(' ')[0];
   const errorMessage = messagePiece;
 


app/cdap/components/Replicator/utilities/index.ts
```diff
@@ -446,7 +446,12 @@ export function identifyReplicatorEntityFromMetadata(metadata: {
  */
 export function parseErrorMessageForTransformations(message: string) {
   const [_, tablePiece, columnPiece, messagePiece] = message.split(' : ');
-  const table = tablePiece.split(' ')[0];
+  // for mssql source, table name will be in format of schema.table
```
de-lan, line 449: Is this fix general or only for MSSQL?

sumengwang: both cases will return just the tablename, if it doesn't have '.' the split will just return itself. I assume tablename should not contain '.'

de-lan (approved):
lgtm

app/cdap/components/Replicator/utilities/index.ts
seanfeld: Yeah split on `.` returnst he orignal string if it doesn't have a `.` .

seanfeld (approved):

PR #448, opened 2022-04-20
[CDAP-18338] fix plugin widgets not showing default values on toggle

sumengwang:
# [CDAP-18338] fix plugin widgets not showing default values on toggle

## Description
Because some widgets conditions are depending on some default values, those values are lost in previous implementation when toggling/switching stuff, change the flow to continuously update values until not able to. 

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18338](https://cdap.atlassian.net/browse/CDAP-18338)

## Test Plan
Manual testing

[CDAP-18338]: https://cdap.atlassian.net/browse/CDAP-18338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/shared/ConfigurationGroup/index.tsx
+++ b/app/cdap/components/shared/ConfigurationGroup/index.tsx
@@ -119,7 +119,7 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
 
   function updateFilteredConfigurationGroup(configGroup, newValues) {
     let newFilteredConfigurationGroup;
-
+    let updatedValuesWithDefault;
     try {
       newFilteredConfigurationGroup = filterByCondition(
         configGroup,
@@ -127,26 +127,45 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
         pluginProperties,
         newValues
       );
+      updatedValuesWithDefault = addDefaultValueToShownProperties(
+        newFilteredConfigurationGroup,
+        newValues
+      );
     } catch (e) {
       newFilteredConfigurationGroup = configGroup;
+      updatedValuesWithDefault = newValues;
       // tslint:disable:no-console
       console.log('Issue with applying filters: ', e);
     }
-
     referenceValueForUnMount.current = {
       configurationGroups: newFilteredConfigurationGroup,
-      values: newValues,
+      values: updatedValuesWithDefault,
     };
     setFilteredConfigurationGroups(newFilteredConfigurationGroup);
+    // to trigger another update if properties are updated with default values
+    changeParentHandler(updatedValuesWithDefault);
     getOrphanedErrors();
   }
-
+  // Some fields may have been cleared and lost their default value
+  // this function adds default value to property which shows and
+  // currently has a null value and non-null default value
+  function addDefaultValueToShownProperties(configGroup, newValues) {
+    const cpyValues = { ...newValues };
+    configGroup.forEach((config) => {
+      config.properties.forEach((property) => {
+        if (property.show && !newValues[property.name] && property.defaultValue) {
+          cpyValues[property.name] = property.defaultValue;
+        }
+      });
+    });
+    return cpyValues;
+  }
   // Watch for changes in values to determine dynamic widget
   useEffect(() => {
-    if (!configurationGroups || configurationGroups.length === 0) {
+    if (!filteredConfigurationGroups || filteredConfigurationGroups.length === 0) {
       return;
     }
-    updateFilteredConfigurationGroup(configurationGroups, values);
+    updateFilteredConfigurationGroup(filteredConfigurationGroups, values);
   }, [values]);
 
   function handleValueChanges(changedValues, params: { [key: string]: boolean } = {}) {
@@ -155,7 +174,10 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
       fcg = filterByCondition(configurationGroups, widgetJson, pluginProperties, changedValues);
     }
 
-    const updatedFilteredValues = removeFilteredProperties(changedValues, fcg);
+    const updatedFilteredValues = addDefaultValueToShownProperties(
+      filteredConfigurationGroups,
+      removeFilteredProperties(changedValues, fcg)
+    );
     changeParentHandler(updatedFilteredValues);
   }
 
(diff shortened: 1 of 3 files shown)


app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -140,7 +146,27 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
+          newValues[property.name] = property.defaultValue;
+          isUpdated = true;
+        }
+      });
+    });
+    // some field have chaining conditions, so updated values may
+    // change some conditions result, hence continuously update
+    // (Is it going to create infinite loop?)
+    if (isUpdated) {
+      updateFilteredConfigurationGroup(configGroup, newValues);
```
seanfeld: you're calling the same function this function is called from if its updated? this logic is super weird to me, couldn't this create an infinite loop?

seanfeld (changes requested):
Am I reading it wrong? it seems like that function has the potential to create an infinite loop?

app/cdap/components/shared/ConfigurationGroup/index.tsx
sumengwang: I thought about that. But the case seems like I need to continuously update it till I cannot do it anymore. I can't think of anything other than recursion.

app/cdap/components/shared/ConfigurationGroup/utilities/index.ts
```diff
@@ -389,7 +389,8 @@ export function removeFilteredProperties(values, filteredConfigurationGroups) {
     filteredConfigurationGroups.forEach((group) => {
       group.properties.forEach((property) => {
         if (property.show === false) {
-          delete newValues[property.name];
+          // instead of deleting the field, set it to null
```
sumengwang: Do we know why it wants to delete it? One of the unit tests specifically wants it to be deleted

app/cdap/components/shared/ConfigurationGroup/index.tsx
sumengwang: Yea I encountered infinite loop cases... I need to think about it more

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -153,7 +153,7 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
     let isUpdated = false;
     configGroup.map((config) => {
       config.properties.map((property) => {
-        if (property.show && newValues[property.name] == null) {
+        if (property.show && newValues[property.name] == null && property.defaultValue != null) {
```
de-lan: `===` and `!==` ?

sumengwang: The value can be undefined or null, this way both can be true

de-lan:
After playing a bit in UI, I suspect that the root cause is `serviceAccountType` is missed in the `updatedValues` passed into parent onChange handler. After we toggle the useConnection twice, only the boolean `useConnection` is passed in and the default serviceAccountType is missing

sumengwang:
> After playing a bit in UI, I suspect that the root cause is `serviceAccountType` is missed in the `updatedValues` passed into parent hon change handler. After we toggle the useConnection twice, only the boolean `useConnection` is passed in and the default serviceAccountType is missed

yes, that's why I made the change to the utilities index file. But in addition to that, I also need to assign the default value to property that has a non-null default value

app/cdap/components/shared/ConfigurationGroup/index.tsx
de-lan: Then we could replace with `!newValues[property.name]`?
We get `if (property.show && !newValues[property.name] && property.defaultValue)`
I'd recommend not using `==` and `!=`, sometimes it introduces implicit bugs

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -119,34 +119,50 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
+          newValues2[property.name] = property.defaultValue;
+        }
+      });
+    });
+    return newValues2;
+  }
   // Watch for changes in values to determine dynamic widget
   useEffect(() => {
-    if (!configurationGroups || configurationGroups.length === 0) {
+    if (!filteredConfigurationGroups || filteredConfigurationGroups.length === 0) {
```
de-lan, line 165: Any reason we want to use `filteredConfigurationGroups` here?

de-lan (approved):
LGTM other than one hanging question

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -119,34 +119,50 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
+    changeParentHandler(newValues2);
     getOrphanedErrors();
   }
-
+  // Some fields may have been cleared and lost their default value
+  // this function adds default value to property which shows and
+  // currently has a null value
+  function addDefaultValueToShownProperties(configGroup, newValues) {
+    const newValues2 = { ...newValues };
+    configGroup.map((config) => {
```
seanfeld: forEach

seanfeld: and the line below

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -119,34 +119,50 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
 
   function updateFilteredConfigurationGroup(configGroup, newValues) {
     let newFilteredConfigurationGroup;
-
+    let newValues2;
```
seanfeld: can we get a better name for this variable

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -119,34 +119,50 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
+    // to trigger another update if newValues2 is different from newValues
+    changeParentHandler(newValues2);
     getOrphanedErrors();
   }
-
+  // Some fields may have been cleared and lost their default value
+  // this function adds default value to property which shows and
+  // currently has a null value
+  function addDefaultValueToShownProperties(configGroup, newValues) {
+    const newValues2 = { ...newValues };
```
seanfeld: and this variable

seanfeld (approved):
nits about variable names and .forEach instead of .map

PR #461, opened 2022-04-25
CDAP-19073: fix the pipeline link issue

de-lan:
# Fix the pipeline link anchor tag issue

## Description
The Anchor tab inside pipeline trigger tabs is not working

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19073](https://cdap.atlassian.net/browse/CDAP-19073)

## Test Plan
Manual

### DIFF
--- a/app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
+++ b/app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
@@ -45,6 +45,10 @@ const useStyle = makeStyle((theme) => {
   };
 });
 
+interface valueProp {
+  required?: boolean;
+}
+
 export function ConnectionConfigForm({
   connectorWidgetJSON,
   connectorProperties,
@@ -68,7 +72,7 @@ export function ConnectionConfigForm({
     const fieldErrors = {};
     for (const [fieldName, value] of Object.entries(connectorProperties)) {
       if (
-        !!value.required &&
+        !!(value as valueProp).required &&
         (!values.hasOwnProperty(fieldName) || values[fieldName] === '' || values[fieldName] == null)
       ) {
         fieldErrors[fieldName] = [
--- a/app/cdap/components/PipelineTriggers/EnabledTriggersTab/EnabledGroupTriggerRow.tsx
+++ b/app/cdap/components/PipelineTriggers/EnabledTriggersTab/EnabledGroupTriggerRow.tsx
@@ -23,11 +23,7 @@ import {
   getGroupPipelineInfo,
 } from 'components/PipelineTriggers/store/PipelineTriggersActionCreator';
 import T from 'i18n-react';
-import {
-  ISchedule,
-  IGroupTrigger,
-  ITriggeringPipelineId,
-} from 'components/PipelineTriggers/store/ScheduleTypes';
+import { ISchedule, IGroupTrigger } from 'components/PipelineTriggers/store/ScheduleTypes';
 import ConfirmationModal from 'components/shared/ConfirmationModal';
 import {
   PipelineName,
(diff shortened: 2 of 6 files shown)


app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
```diff
@@ -67,10 +67,7 @@ export function ConnectionConfigForm({
   const onCreate = () => {
     const fieldErrors = {};
     for (const [fieldName, value] of Object.entries(connectorProperties)) {
-      if (
-        !!value.required &&
-        (!values.hasOwnProperty(fieldName) || values[fieldName] === '' || values[fieldName] == null)
-      ) {
+      if ((value as any).required && (!values.hasOwnProperty(fieldName) || !values[fieldName])) {
```
de-lan: The build is reporting error: `Property 'required' does not exist on type 'unknown'.`

Also `values[fieldName] === '' || values[fieldName] == null` is basically equal to `!values[fieldName]`

de-lan: My bad, I missed the case that `values[fieldName] = false` and they don't equal in this corner case. Updated

mrahanjam: The first error makes sense in the eyes of TS compiler. No guarantee a type `any` has a `required` prop. But ~for your second statement, I'm guessing we want to avoid treating `values[fieldName] === 0` as `false`~

EDIT: saw your second comment too late

de-lan: You also made a nice counter example, thanks!

app/cdap/components/PipelineTriggers/store/PipelineTriggersActionCreator.ts
```diff
@@ -636,3 +636,11 @@ function _transformSchedule(existingSchedules: ISchedule[]) {
 
   return [...transformedSchedules, ...enabledGroupSchedule];
 }
+
+/**
+ * Method to transform the single trigger to OR group trigger.
+ * @param link - the link to route to.
+ */
+export const handlePipelineLinkClick = (link: string) => {
```
mrahanjam: I'd move this to `services/helpers` or the `shared` directory as it's not really creating an action in the store (no calls to `PipelineTriggersStore` or state changes). It is also being used by more than one component. You could also use a generic name such as `openLinkInNewTab` or something similar. We might end up using it in other places as well.

de-lan: Good idea, will update

app/cdap/components/Connections/Create/ConnectionConfiguration/ConnectionConfigForm.tsx
```diff
@@ -68,7 +68,7 @@ export function ConnectionConfigForm({
     const fieldErrors = {};
     for (const [fieldName, value] of Object.entries(connectorProperties)) {
       if (
-        !!value.required &&
+        !!(value as any).required &&
```
mrahanjam: I think if you use `value.hasOwnProperty(required) && value.required`, you won't get the TS compiler error. Haven't tested it though

de-lan: image
hmmm, it still shows up

mrahanjam (approved):
LGTM! If you really wanna get rid of that TS error, you can define an interface for value and have required as an optional prop like:

```javascript
interface valueProp {
    required?: boolean
}
```

PR #462, opened 2022-04-25
[CDAP-19137|18338] Address new bugs for 18338

sumengwang:
# [CDAP-19137|18338] Address new bugs for 18338

## Description
[pull 448](https://github.com/cdapio/cdap-ui/pull/448) caused an infinite useEffect loop. Because the value it checks is an object. Due to the design of useEffect, as long as a new object is set, even if the contents are the same, useEffect will still trigger, hence infinite loop. The fix is to use "isEqual" from lodash to decide whether to set new object.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19137](https://cdap.atlassian.net/browse/CDAP-19137)
[CDAP-18338](https://cdap.atlassian.net/browse/CDAP-18338)

## Test Plan
Probably should write a new e2e test
pipeline.spec.ts and pipeline.postrunactions.spec.ts now pass (previously failed in e2e)

### DIFF
--- a/app/cdap/components/shared/ConfigurationGroup/index.tsx
+++ b/app/cdap/components/shared/ConfigurationGroup/index.tsx
@@ -18,7 +18,11 @@ import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import withStyles, { WithStyles, StyleRules } from '@material-ui/core/styles/withStyles';
 import { IWidgetJson, PluginProperties } from './types';
-import { processConfigurationGroups, removeFilteredProperties } from './utilities';
+import {
+  processConfigurationGroups,
+  removeFilteredProperties,
+  addDefaultValueToShownProperties,
+} from './utilities';
 import { objectQuery } from 'services/helpers';
 import { useOnUnmount } from 'services/react/customHooks/useOnUnmount';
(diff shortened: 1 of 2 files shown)


app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -138,6 +143,11 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
       values: newValues,
     };
     setFilteredConfigurationGroups(newFilteredConfigurationGroup);
+    // to trigger another update if properties are updated with default values
+    // only trigger if the contents of two objs are different
+    if (!isEqual(newValues, updatedValuesWithDefault)) {
+      changeParentHandler(updatedValuesWithDefault);
```
de-lan, line 153: Looks like we missed some code? 
image

de-lan (approved):
LGTM, but maybe need another approval as we discussed last week?

Also on nitpick is that please make the commit message more informative.

seanfeld:
should probably drop this pr?

PR #466, opened 2022-04-27
Use Selenium and Github Workflows for E2E tests

mrahanjam:
# Use Selenium and Github Workflows for E2E tests

## Description
This PR creates the github action/workflow to run the integration tests with selenium, it also includes a dummy test.

## PR Type
- [ ] Bug Fix
- [X] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19260](https://cdap.atlassian.net/browse/CDAP-19260)

## Test Plan
N/A

## Screenshots
N/A


selenium/integration/starter.spec.ts
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
seanfeld: is this a moved file? 2022

seanfeld (approved):
looks fine but i'm curious why you updated the node version for the test runner? I think its probably fine since its test only but we can't update node leswhere unfortunately :(

mrahanjam:
> looks fine but i'm curious why you updated the node version for the test runner? I think its probably fine since its test only but we can't update node leswhere unfortunately :(

I think the main reason is to start the sandbox in the test env. It complains with node versions smaller than 12. ~I borrowed that part from the old workflow file.~ I believe 12 has better compatibility with the selenium-webdriver as well.

selenium/integration/starter.spec.ts
mrahanjam: will update to 2022

selenium/integration/starter.spec.ts
```diff
@@ -0,0 +1,45 @@
+/*
```
mrahanjam, line 1: This is a dummy test to make sure the webdriver works in the test env. Will be removed once we have real tests.

.github/workflows/node.js.yml
```diff
@@ -36,14 +41,30 @@ jobs:
...
 
-    - name: Run Unit Tests
-      run: yarn run jest
+    - name: Start CDAP-UI and run the tests
+      run: |
+        Xvfb :1 -screen 0 1024x768x24 & export DISPLAY=:1 # Set Display port
+        yarn run cdap-full-build-more-memory # Build UI
+        yarn start & > server.out
+        yarn run test:unit
+        yarn run test:integration
```
njbriggs: Nit: newline before EOF

njbriggs: And a couple other files - please double-check

njbriggs (approved):

PR #467, opened 2022-04-27
[CDAP-18338] fix configForm ui not remembering user input on toggle

sumengwang:
# [CDAP-18338] fix configForm ui not remembering user input on toggle

## Description
Changed the approach to use another state to remember the current user input and insert the values back to the new filtered values after toggling. 

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18338](https://cdap.atlassian.net/browse/CDAP-18338)

[CDAP-18338]: https://cdap.atlassian.net/browse/CDAP-18338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/shared/ConfigurationGroup/index.tsx
+++ b/app/cdap/components/shared/ConfigurationGroup/index.tsx
@@ -18,7 +18,12 @@ import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import withStyles, { WithStyles, StyleRules } from '@material-ui/core/styles/withStyles';
 import { IWidgetJson, PluginProperties } from './types';
-import { processConfigurationGroups, removeFilteredProperties } from './utilities';
+import {
+  addCurrentValueToShownProperty,
+  processConfigurationGroups,
+  removeFilteredProperties,
+  replaceDifferenceInObjects,
+} from './utilities';
 import { objectQuery } from 'services/helpers';
 import { useOnUnmount } from 'services/react/customHooks/useOnUnmount';
 import defaults from 'lodash/defaults';
@@ -85,6 +90,8 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
   }>({});
   const [filteredConfigurationGroups, setFilteredConfigurationGroups] = useState([]);
   const [orphanErrors, setOrphanErrors] = useState([]);
+  // a state to remember the properties values after each user interaction
+  const [currentValues, setCurrentValues] = useState(values);
 
   // Initialize the configurationGroups based on widgetJson and pluginProperties obtained from backend
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


app/cdap/components/shared/ConfigurationGroup/utilities/index.ts
```diff
@@ -397,3 +397,26 @@ export function removeFilteredProperties(values, filteredConfigurationGroups) {
 
   return newValues;
 }
+
+/**
+ * replace the values in values1 with values2
```
seanfeld, line 402: Can you write jsdoc format comments here - add @param values (don't include types - let typescript handle types) like the other functions in this file?

app/cdap/components/shared/ConfigurationGroup/utilities/index.ts
```diff
@@ -397,3 +397,26 @@ export function removeFilteredProperties(values, filteredConfigurationGroups) {
...
+ */
+export function replaceDifferenceInObjects(values1, values2) {
+  Object.keys(values2).forEach((key) => {
+    values1[key] = values2[key];
+  });
+}
+
+/**
```
seanfeld, line 415: here too

seanfeld (changes requested):

seanfeld (approved):
@mrahanjam can you check this out

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -85,6 +90,8 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
   }>({});
   const [filteredConfigurationGroups, setFilteredConfigurationGroups] = useState([]);
   const [orphanErrors, setOrphanErrors] = useState([]);
+  // a state to remember the properties values after each user interaction
+  const [currentValues, setCurrentValues] = useState(values);
```
de-lan, line 94: Seems like we're never calling `setCurrentValues`? Somehow I have a mixed feeling...

sumengwang: Yea I chose to directly modify currentValues since `setCurrentValues` is async. Maybe there's a better way

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -154,7 +164,8 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
     if (params.updateFilteredConfigurationGroups) {
       fcg = filterByCondition(configurationGroups, widgetJson, pluginProperties, changedValues);
     }
-
+    // update currentValues state with whatever changed
+    replaceDifferenceInObjects(currentValues, changedValues);
```
de-lan: Referring to [this](https://github.com/cdapio/cdap-ui/pull/467/files#r860314667), should we simply use `setCurrentValues(changedValues)`?

mrahanjam:
> @mrahanjam can you check this out

Sure, I'll review it by EOD today

app/cdap/components/shared/ConfigurationGroup/index.tsx
sumengwang: `currentValues` have all the keys, `changedValues` may contain partial keys since some will get deleted in previous `removeFilteredProperties` operation. Also it's async so I'm worried when I want to use the `currentValues` later, it's not getting the updated value

mrahanjam: Give this a try and see if it works. We probably don't want to mutate our state object.
```javascript
setCurrentValues(changedValues)
useEffect(() => {
    const updatedFilteredValues = removeFilteredProperties(changedValues, fcg);
    changeParentHandler(updatedFilteredValues);
  }, [currentValues]);
```

sumengwang: turns out using setCurrentValues is fine

de-lan (approved):

PR #492, opened 2022-05-06
handle ts type errors and warnings in UI build

mrahanjam:
# Handle TS type errors and warnings in UI build

## Description
This PR handles the two TS type errors I kept seeing while building the UI. They are not the ideal solutions but it's pretty much the best I could do at this point. 
You can read more about the `button` issue [here](https://github.com/mui/material-ui/issues/14971). If/when we upgrade to MUI v5, we can use `ListItemButton` instead of type `any` for the `button` prop.
For the second one, it's hard to come up with an interface for the event obj as it depends on the props of the element it's coming from, so I just decided to give it a type `any`.

Edit: Decided to handle warnings as well

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [X] General Improvement
- [ ] Cherry Pick

## Links
N/A

## Test Plan
N/A

## Screenshots
N/A

### DIFF
--- a/app/cdap/components/Connections/Create/CategorizedConnectors/ActiveTab.tsx
+++ b/app/cdap/components/Connections/Create/CategorizedConnectors/ActiveTab.tsx
@@ -179,9 +179,9 @@ export function ActiveConnectionTab({ connector, onConnectorSelection, search, o
                     <ListItem
                       divider
                       disableGutters
-                      button={!hasOlderVersion}
+                      button={!hasOlderVersion as any}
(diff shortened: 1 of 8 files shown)


sumengwang (approved):
Nice, my vscode finally stops complaining

de-lan (approved):

seanfeld (approved):

sumengwang (approved):

PR #493, opened 2022-05-06
CDAP-19195: fix, add a ClickAwayListener to Select widget so that any…

de-lan:
Add a ClickAwayListener to Select component widget so that any click-away event will close the dropdown menu

# TITLE

## Description
The Select component should automatically detect an outside Click event and collapse the dropdown menu. However not sure why the clicks on top header bar is not triggering it. Here I manually added a `ClickAwayListener`. Feels like a bit overkill and suggestions are welcome!

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19195](https://cdap.atlassian.net/browse/CDAP-19195)

### DIFF
--- a/app/cdap/components/AbstractWidget/FormInputs/Select/index.tsx
+++ b/app/cdap/components/AbstractWidget/FormInputs/Select/index.tsx
@@ -18,12 +18,13 @@ import { IWidgetProps } from 'components/AbstractWidget';
 import Input from '@material-ui/core/Input';
 import InputBase from '@material-ui/core/InputBase';
 import MenuItem from '@material-ui/core/MenuItem';
-import React from 'react';
+import React, { useState } from 'react';
 import Tooltip from '@material-ui/core/Tooltip';
 import { WIDGET_PROPTYPES } from 'components/AbstractWidget/constants';
 import { isNilOrEmptyString } from 'services/helpers';
 import { objectQuery } from 'services/helpers';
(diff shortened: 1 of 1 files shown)


njbriggs:
This does feel like overkill to me as well. Can you spend some more time and see if you can figure out why clicking on the header is not closing the select?

seanfeld:
Maybe there is a event.preventDefault(); or something like that that is preventing the event from bubbling up?

de-lan:
> This does feel like overkill to me as well. Can you spend some more time and see if you can figure out why clicking on the header is not closing the select?

Yeah I'm looking deeper. However if there's no better solution before end of this week(this code freeze), I'll probably merge it as a temp approach

de-lan:
> Maybe there is a event.preventDefault(); or something like that that is preventing the event from bubbling up?

Will take a deeper look

njbriggs:
This is not marked for 6.7.0 - it's been moved to 6.7.1.

seanfeld:
i found an easy fix will make a pr

seanfeld:
https://github.com/cdapio/cdap-ui/pull/497 this fixes it

de-lan:
Closing since there's a better way using z-index

PR #494, opened 2022-05-06
[CDAP-18893] use display-name property whenever possible

sumengwang:
# [CDAP-18893] use display-name property whenever possible

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18893](https://cdap.atlassian.net/browse/CDAP-18893)

## Screenshots
image
image

[CDAP-18893]: https://cdap.atlassian.net/browse/CDAP-18893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/Connections/Create/reducer.ts
+++ b/app/cdap/components/Connections/Create/reducer.ts
@@ -296,14 +296,36 @@ function getArtifactKey(artifact) {
   return `${artifact.name}-${artifact.version}-${artifact.scope}`;
 }
 
-function getMapOfConnectorToPluginProperties(allConnectors) {
+export function getMapOfConnectorToPluginProperties(allConnectors) {
   const mapOfConnectorPluginProperties = {};
   for (const connectorProps of allConnectors) {
     const pluginKey = getPluginKey(connectorProps.properties);
+    const widgetJSONKey = `widgets.${pluginKey}`;
+    const parsedWidgetJson = JSON.parse(connectorProps.properties[widgetJSONKey]);
     const artifactKey = getArtifactKey(connectorProps);
-    mapOfConnectorPluginProperties[`${pluginKey}-${artifactKey}`] = connectorProps.properties;
+    const parsedWidgetJsonKey = `parsedWidgetJson.${pluginKey}`;
+    mapOfConnectorPluginProperties[`${pluginKey}-${artifactKey}`] = {
+      ...connectorProps.properties,
+      [parsedWidgetJsonKey]: parsedWidgetJson,
+    };
   }
   return mapOfConnectorPluginProperties;
+}
+
+export function getSelectedConnectorDisplayName(selectedConnector, mapOfConnectorPluginProperties) {
+  if (!selectedConnector) {
+    return;
+  }
+  const pluginKey = `${selectedConnector.name}-${selectedConnector.type}`;
+  const artifactKey = getArtifactKey(selectedConnector.artifact);
+  const parsedwidgetJSONKey = `parsedWidgetJson.${pluginKey}`;
+  const selectedWidgetJSON =
+    mapOfConnectorPluginProperties[`${pluginKey}-${artifactKey}`][parsedwidgetJSONKey];
+
+  if (selectedWidgetJSON['display-name']) {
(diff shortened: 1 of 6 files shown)


seanfeld (approved):

njbriggs:
I think this causes a mismatch with the listing:
image

We will need this change eventually, but I think we should coordinate it with the change to the listing so they are in sync.

sumengwang:
> I think this causes a mismatch with the listing
> We will need this change eventually, but I think we should coordinate it with the change to the listing so they are in sync.

I see your point. Right now it only fetches the widget json when you click one connector.

de-lan (approved):

sumengwang:
Right now all places have consistent display name
image
image

app/cdap/components/Connections/Create/reducer.ts
```diff
@@ -303,6 +303,30 @@ function getMapOfConnectorToPluginProperties(allConnectors) {
   return mapOfConnectorPluginProperties;
 }
 
+export function addDisplayNameToConnector(connectors, allConnectorsPluginProperties) {
+  connectors.forEach((connector) => {
+    allConnectorsPluginProperties.forEach((connectorPluginProperties) => {
```
njbriggs: Can you use `mapOfConnectorPluginProperties` and avoid the nested loops here?

app/cdap/components/Connections/Create/reducer.ts
```diff
@@ -303,6 +303,30 @@ function getMapOfConnectorToPluginProperties(allConnectors) {
...
+        connectorPluginProperties.version === connector.artifact.version &&
+        connectorPluginProperties.scope === connector.artifact.scope
+      ) {
+        // add display-name to connector
+        const pluginKey = `${connector.name}-connector`;
+        const widgetJSONKey = `widgets.${pluginKey}`;
+        if (widgetJSONKey in connectorPluginProperties.properties) {
+          const widgetJSON = JSON.parse(connectorPluginProperties.properties[widgetJSONKey]);
+          if (widgetJSON['display-name']) {
+            connector.displayName = widgetJSON['display-name'];
```
njbriggs: Rather than adding a field here, it would probably make more sense to pass down `mapOfConnectorPluginProperties` to the component doing the rendering. This will help us define specific types for the store if we need to.

sumengwang: It seems straightforward to use `mapOfConnectorPluginProperties` in Create/index, but SidePanel/index is not using the reducer. What's the best way to go about it?

sumengwang: addressed the comment

app/cdap/components/Connections/Create/reducer.ts
```diff
@@ -303,6 +303,22 @@ function getMapOfConnectorToPluginProperties(allConnectors) {
...
 }
 
+export function getSelectedConnectorDisplayName(selectedConnector, mapOfConnectorPluginProperties) {
+  if (!selectedConnector) {
+    return;
+  }
+  const pluginKey = `${selectedConnector.name}-${selectedConnector.type}`;
+  const artifactKey = getArtifactKey(selectedConnector.artifact);
+  const widgetJSONKey = `widgets.${pluginKey}`;
+  const selectedWidgetJSON = JSON.parse(
```
njbriggs: Is there any way to parse this earlier and store the result? My concern is that this requires parsing the JSON for every name appearance on every render.

njbriggs:
Overall, this approach looks great! I had 1 piece of feedback. Also, I think it needs a rebase.

app/cdap/components/Connections/Create/reducer.ts
sumengwang: Now I save the parsed Json into the mapOfConnectorPluginProperties, let me know if this aligns with your thought

njbriggs (approved):
LGTM!

PR #497, opened 2022-05-11
Change z index of layer so that it sits above everything on page for select

seanfeld:
# CDAP-19195

## Description
Change z index of select layer so that it sits above things on the page - I think this is from us making our own modal and using z index to accomplish it.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19195](https://cdap.atlassian.net/browse/CDAP-19195)

### DIFF
--- a/app/cdap/components/AbstractWidget/FormInputs/Select/index.tsx
+++ b/app/cdap/components/AbstractWidget/FormInputs/Select/index.tsx
@@ -145,6 +145,9 @@ const CustomSelect: React.FC<ISelectProps> = ({
       }}
       MenuProps={{
         getContentAnchorEl: null,
+        style: {
+          zIndex: 1302,
+        },
         anchorOrigin: {
           vertical: 'bottom',
           horizontal: 'left',


de-lan (approved):
Verified it's working

app/cdap/components/AbstractWidget/FormInputs/Select/index.tsx
```diff
@@ -145,6 +145,9 @@ const CustomSelect: React.FC<ISelectProps> = ({
       }}
       MenuProps={{
         getContentAnchorEl: null,
+        style: {
+          zIndex: 1302,
```
mrahanjam, line 149: nit: can we do 999? feel free to ignore.

mrahanjam (approved):

app/cdap/components/AbstractWidget/FormInputs/Select/index.tsx
seanfeld: for whatever reason our code uses 1301 in certain places, I think _because_ material ui uses this stuff.. I hesitate to push it any higher *AND* I would also recommend we don't push this into 6.7 because it there is a small chance it might cause a worse bug elsewhere.

mrahanjam: gotcha! looks good to me then 👍

sumengwang (approved):

njbriggs (approved):
You might need to rebase this to get the new build command.

PR #498, opened 2022-05-11
remove chromedriver as a dependency

mrahanjam:
# Fix build issue

## Description
This PR removes chromedriver from the package,json which is incompatible with node 10.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [X] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.22.7]
+        node-version: [10.16]
         java-version: [8]
 
     # Cancel ongoing runs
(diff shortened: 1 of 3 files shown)


mrahanjam:
It looks like the github ubuntu image has chromedriver installed by default, there's also a way to install it manually through github actions, but I want to resort to that when tests fail. SO far looks like selenium tests can run without an issue.

seanfeld (approved):
does your local yarn build with this?

njbriggs (approved):

mrahanjam:
> does your local yarn build with this?

yessir

PR #504, opened 2022-05-12
Make Advanced config group collapsible

mrahanjam:
# CDAP-18541: Make config groups collapsible

## Description
This PR suggests a way for config groups to become collapsible.

## PR Type
- [ ] Bug Fix
- [X] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
[CDAP-18541](https://cdap.atlassian.net/browse/CDAP-18541)

## Test Plan
Manual

## Screenshots

https://user-images.githubusercontent.com/94018249/167997388-09ac2f09-5832-465b-bb52-009937d840ca.mov

### DIFF
--- a/app/cdap/components/shared/ConfigurationGroup/index.tsx
+++ b/app/cdap/components/shared/ConfigurationGroup/index.tsx
@@ -36,6 +36,7 @@ import {
 } from 'components/shared/ConfigurationGroup/utilities/DynamicPluginFilters';
 import { IErrorObj } from 'components/shared/ConfigurationGroup/utilities';
 import { h2Styles } from 'components/shared/Markdown/MarkdownHeading';
+import AccordionWrapper from './AccordionWrapper';
 
 const styles = (theme): StyleRules => {
   return {
@@ -106,7 +107,6 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
       widgetConfigurationGroup,
       widgetOutputs
     );
-
     setConfigurationGroups(processedConfigurationGroup.configurationGroups);
 
     // We don't need to add default values for plugins in published pipeline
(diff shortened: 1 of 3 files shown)


app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -37,6 +38,8 @@ import {
 import { IErrorObj } from 'components/shared/ConfigurationGroup/utilities';
 import { h2Styles } from 'components/shared/Markdown/MarkdownHeading';
 
+const COLLAPSIBLE_CONFIG_GROUPS = ['Advanced'];
```
mrahanjam: This gives us the ability to add any other config group to this list and make them collapsible. We could also add a `collapsible` prop to the plugins' json files, and use that to populate `initialConfigGroupExpandedState` on line 118, and also use it when rendering further down instead of line 274. I kinda like to keep the control here on the UI side but the only downside I can see is us relying on the `label` from the plugin's json file which is probably not too risky.

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -252,50 +271,61 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
         if (group.show === false) {
           return null;
         }
 
         return (
           <div key={`${group.label}-${i}`} className={classes.group}>
-            <div className={classes.groupTitle}>
-              <h2 className={classes.h2Title}>{group.label}</h2>
+            <div className={classes.groupTitle && isCollapsible && classes.pointer}>
+              <h2 className={classes.h2Title} onClick={() => toggleConfGroup(group.label)}>
```
njbriggs: Did you look into Mui Accordion?

app/cdap/components/shared/ConfigurationGroup/index.tsx
bdmogal: yeah, this makes it inflexible. Providing control in the widget JSON with something like ``hideByDefault: true|false`` would be better. That way, any group could use it.

app/cdap/components/shared/ConfigurationGroup/index.tsx
mrahanjam: I have not, but I will

app/cdap/components/shared/ConfigurationGroup/index.tsx
mrahanjam: That's true but if we want to get this out for 6.7, I don't think we can apply the changes to the widget JSONs unless we push this back

avirmani: How big is that change to add the hidebyDefault and let widget jsons define it?

mrahanjam: On the UI it's not much of a change. I just make a small change in my PR.

app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -239,6 +239,53 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
...
+            <PropertyRow
+              key={`${property.name}-${j}`}
+              widgetProperty={property}
+              pluginProperty={pluginProperties[property.name]}
+              value={values[property.name]}
+              onChange={handleValueChanges}
+              extraConfig={extraConfig}
+              disabled={disabled}
+              locked={lockedProperties && lockedProperties[property.name]}
+              errors={errorObjs}
```
njbriggs, line 281: For future: If there's a highlighted error field in the hidden section, can we auto-expand it?

njbriggs (approved):
Please squash

app/cdap/components/shared/ConfigurationGroup/index.tsx
mrahanjam: sure, will take a note

seanfeld (approved):

PR #510, opened 2022-05-13
CDAP-19219: convert admin.spec.ts to selenium

mrahanjam:
# Convert admin.spec.ts to selenium

## Description
This PR converts `admin.spec.ts` test from cypress to selenium.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [X] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19219](https://cdap.atlassian.net/browse/CDAP-19219)

## Test Plan
Automated

### DIFF
--- a/app/cdap/components/Administration/AdminConfigTabContent/SystemPrefsAccordion/index.js
+++ b/app/cdap/components/Administration/AdminConfigTabContent/SystemPrefsAccordion/index.js
@@ -185,6 +185,7 @@ export default class SystemPrefsAccordion extends Component {
           className="btn btn-secondary"
           onClick={this.togglePrefsModal}
           data-cy="edit-system-prefs-btn"
+          data-testid="edit-system-prefs-btn"
         >
           {T.translate(`${PREFIX}.create`)}
         </button>
@@ -220,6 +221,7 @@ export default class SystemPrefsAccordion extends Component {
           expanded: this.props.expanded,
         })}
         data-cy="system-prefs-accordion"
+        data-testid="system-prefs-accordion"
       >
         {this.renderLabel()}
         {this.renderContent()}
--- a/package.json
+++ b/package.json
@@ -216,6 +216,7 @@
     "mousetrap": "1.6.3",
     "natural-orderby": "2.0.3",
     "ngreact": "0.5.2",
+    "node-fetch": "2",
     "numeral": "2.0.6",
     "popper.js": "^1.16.0",
     "prismjs": "1.17.1",
--- a/selenium/integration/admin.spec.ts
+++ b/selenium/integration/admin.spec.ts
@@ -1,0 +1,61 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License'); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import { By } from 'selenium-webdriver';
+import { dataTestId, buildChromeDriver, loginIfRequired } from '../support/utils';
+import { CONFIGURATION_URL } from '../support/constants';
+
+let driver;
+const TEST_KEY = 'name';
+const TEST_VALUE = 'hello';
+
+describe('Setting and saving preferences', () => {
+  beforeAll(async () => {
+    driver = buildChromeDriver();
+    await loginIfRequired(driver);
+  });
+
+  it('Should show error message if user tries to set profile at the instance level', async () => {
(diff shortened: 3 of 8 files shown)


app/cdap/components/Administration/AdminConfigTabContent/SystemPrefsAccordion/index.js
```diff
@@ -185,6 +185,7 @@ export default class SystemPrefsAccordion extends Component {
           className="btn btn-secondary"
           onClick={this.togglePrefsModal}
           data-cy="edit-system-prefs-btn"
+          data-se="edit-system-prefs-btn"
```
mrahanjam: Should we consider a generic name for this attribute? I couldn't think of a good one other than `data-testid` maybe? open to suggestions.

package.json
```diff
@@ -216,6 +216,7 @@
     "mousetrap": "1.6.3",
     "natural-orderby": "2.0.3",
     "ngreact": "0.5.2",
+    "node-fetch": "2",
```
mrahanjam, line 219: Axios has some compatibility issues with `jest`, decided to use `node-fetch`.

selenium/support/constants.ts
```diff
@@ -0,0 +1,46 @@
+/*
```
mrahanjam, line 1: Copied over from `Cypress` and added a couple of new ones.

selenium/support/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  options.addArguments('--no-sandbox')
+  // options.addArguments('--headless')
+
+  return new Builder()
+    .forBrowser('chrome')
+    .setChromeOptions(options)
+    .build()
+}
+
+export const loginIfRequired = async (driver) => {
```
mrahanjam, line 40: I wasn't able to fully test this function (particularly the block starting at line 50). Was not able to simulate the login process locally. On paper it should work, but we might see failures when logging in is actually required.

selenium/integration/admin.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import { CONFIGURATION_URL } from '../support/constants';
+
+let driver;
+const TEST_KEY = 'name';
+const TEST_VALUE = 'hello';
+
+describe('Setting and saving preferences', () => {
+  beforeAll(async () => {
+    driver = buildChromeDriver();
+    await loginIfRequired(driver);
```
mrahanjam, line 28: I'm not sure why the cypress test calls this function. it doesn't really use the `CDAP_AUTH_TOKEN` anywhere. Decided to copy it over just in case.

selenium/integration/admin.spec.ts
```diff
@@ -0,0 +1,61 @@
+/*
```
de-lan, line 1: I feel this was well aware in the team but I perhaps didn't yet join at that time. But what's the gain from selenium over cypress in cdap app?

mrahanjam: AFAIK, the main reason is that the backend team started using Selenium in their E2E tests, and we decided to switch to Selenium so that we are all using the same framework.

de-lan: Ah gotcha

mrahanjam:
I may also add that based on my findings on Selenium's webdriver performance, we should try to use the css selector whenever possible (i.e., going down the DOM), and use xpath when looking for parents/ancestors. Css is faster, lighter, and easier to read.

yarn.lock
```diff
@@ -6861,20 +6861,10 @@ caniuse-db@^1.0.30000529, caniuse-db@^1.0.30000634, caniuse-db@^1.0.30000639:
...
-caniuse-lite@^1.0.30001125:
-  version "1.0.30001270"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001270.tgz#cc9c37a4ec5c1a8d616fc7bace902bb053b0cdea"
-  integrity sha512-TcIC7AyNWXhcOmv2KftOl1ShFAaHQYcB/EPL/hEyMrcS7ZX0/DvV1aoy6BzV0+16wTpoAyTMGDNAJfSqS/rz7A==
-
-caniuse-lite@^1.0.30001265:
-  version "1.0.30001269"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001269.tgz#3a71bee03df627364418f9fd31adfc7aa1cc2d56"
-  integrity sha512-UOy8okEVs48MyHYgV+RdW1Oiudl1H6KolybD6ZquD0VcrPSgj25omXO1S7rDydjpqaISCwA8Pyx+jUQKZwWO5w==
+caniuse-lite@^1.0.30001012, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001125, caniuse-lite@^1.0.30001181, caniuse-lite@^1.0.30001265:
```
mrahanjam, line 6864: this will remove the update reminder for `canisue-liter` when building the UI

app/cdap/components/Administration/AdminConfigTabContent/SystemPrefsAccordion/index.js
njbriggs: `data-testid` works for me

njbriggs (approved):

de-lan (approved):

PR #514, opened 2022-05-18
add tethering related tests

mrahanjam:
# Add Tethering Related Tests

## Description
This PR adds the automated test scripts for odf registration tasks on the UI.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [X] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-18887](https://cdap.atlassian.net/browse/CDAP-18887)

## Test Plan
TBD

### DIFF
--- a/app/cdap/components/shared/ConfigurationGroup/WidgetWrapper/index.tsx
+++ b/app/cdap/components/shared/ConfigurationGroup/WidgetWrapper/index.tsx
@@ -142,6 +142,7 @@ const WidgetWrapperView: React.FC<IWidgetWrapperProps> = ({
       onFocus={onFocus}
       onBlur={onBlur}
       data-cy="widget-wrapper-container"
+      data-testid="widget-wrapper-container"
     >
       {!hideWrapper && !hideLabel && (
         <div className={`widget-wrapper-label ${classes.label}`}>
@@ -160,6 +161,7 @@ const WidgetWrapperView: React.FC<IWidgetWrapperProps> = ({
           disabled={disabled}
           errors={errors}
           dataCy={pluginProperty.name}
+          dataTestId={pluginProperty.name}
         />
       </div>
       {!hideDescription && !hideWrapper && (
--- a/jest.config.js
+++ b/jest.config.js
@@ -43,6 +43,7 @@ module.exports = {
   modulePathIgnorePatterns: [
     './app/cdap/components/AbstractWidget/SchemaEditor/Context/__tests__/schemas.js',
     './app/cdap/components/AbstractWidget/SchemaEditor/Context/__tests__/data',
+    './selenium/integration/tethering',
   ],
   modulePaths: ['./app/cdap/'],
   preset: 'ts-jest/presets/js-with-babel',
--- a/selenium/integration/admin.spec.ts
+++ b/selenium/integration/admin.spec.ts
@@ -44,9 +44,16 @@ describe('Setting and saving preferences', () => {
   });
 
   it('Should allow user to save valid preference at instance level after fixing error', async () => {
+    await driver.get(CONFIGURATION_URL);
+    await driver.sleep(1000);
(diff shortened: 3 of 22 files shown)


mrahanjam:
These tests will be ignored in the github actions, I created this PR for basically storing the scripts. Actual running of the tests will be handled separately once the clusters setup for odf is complete. This PR includes a small update to `admin.spec.ts` test as well.

selenium/integration/admin.spec.ts
```diff
@@ -44,9 +44,16 @@ describe('Setting and saving preferences', () => {
   });
 
   it('Should allow user to save valid preference at instance level after fixing error', async () => {
+    await driver.get(CONFIGURATION_URL);
```
mrahanjam, line 47: Decided to load the page again in this test. Saw a couple of failures. Can't trust selenium to pick up where the previous test left off.

jest.config.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
   modulePathIgnorePatterns: [
     './app/cdap/components/AbstractWidget/SchemaEditor/Context/__tests__/schemas.js',
     './app/cdap/components/AbstractWidget/SchemaEditor/Context/__tests__/data',
+    './selenium/integration/tethering',
```
de-lan, line 46: So we're ignoring the tethering tests for now?

mrahanjam: yes they need their own setup (i.e., a cdf instance + an hdf cluster). They'll probably move to their own repo once the test setup is done by the backend team.

app/cdap/components/shared/ConfigurationGroup/WidgetWrapper/index.tsx
```diff
@@ -160,6 +161,7 @@ const WidgetWrapperView: React.FC<IWidgetWrapperProps> = ({
           disabled={disabled}
           errors={errors}
           dataCy={pluginProperty.name}
+          dataTestId={pluginProperty.name}
```
njbriggs, line 164: Should we change components to only accept `dataTestId` now? Then have the component set both `data-cy` and `data-test-id` on the DOM element. Not a big difference but would save a little bit of cleanup later.

selenium/integration/tethering/manage.requests.cdf.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    await driver.sleep(1000);
+  }
+
+  it('Should navigate to tethering page', async () => {
+    await navigateToTetheringPage();
+    const url = await driver.getCurrentUrl();
+    expect(url.toString()).toContain('/tethering');
+  })
+
+  it('Should successfully accept a tethering request', async() => {
```
njbriggs, line 36: It looks like this test case depends on the previous case? Each test case should be runnable without any other case. Logic to load the page should go within the case or in a `beforeEach` block.

de-lan: I agree that each test should be independent. However in the codebase, e.g in pipeline related tests, there are many test cases depending on the precedents...

njbriggs: Understood, but let's not add to that debt.

app/cdap/components/shared/ConfigurationGroup/WidgetWrapper/index.tsx
mrahanjam: Ideally yes we'd want that to reduce cleanup work, but in some cases I don't think it's doable. For example in `app/cdap/components/AbstractWidget/StateWrapper.js`, `dataCy` is passed to various components depending on `Comp` value and removing `dataCy` in one abstract widget component and not the other might start to get confusing/messy. I think removing all usages of `dataCy` in one PR might be a better way to handle it.

njbriggs (approved):
Please squash before merging

de-lan (approved):

PR #519, opened 2022-05-25
feat(CDAP-19229): convert page.level.errors.spec to selenium

de-lan:
## Description
Convert page.level.errors.spec to selenium

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [x] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19229](https://cdap.atlassian.net/browse/CDAP-19229)

## Test Plan
Manual
## Screenshots

image

### DIFF
--- a/selenium/integration/page.level.errors.spec.ts
+++ b/selenium/integration/page.level.errors.spec.ts
@@ -1,0 +1,126 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License'); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import { BASE_URL, CDAP_URL, METADATA_URL, PIPELINES_URL } from '../support/constants';
+import { buildChromeDriver, loginIfRequired, makeElementHelpers } from '../support/utils';
+
+describe('Page level error because of ', () => {
+  let driver;
+  let se;
+  const FAKE_NAMESPACE = 'fakeNamespace';
+  const SELECTOR_404_MSG = 'page-404-error-msg';
+  const SELECTOR_404_DEFAULT_MSG = 'page-404-default-msg';
+
+  beforeAll(async () => {
+    driver = buildChromeDriver();
+    se = makeElementHelpers(driver);
+    await loginIfRequired(driver);
+  });
+
+  it('no namespace in home page should show 404', async () => {
+    // Go to home page
+    await driver.get(`${CDAP_URL}/${FAKE_NAMESPACE}`);
+    await se.waitByTestId(SELECTOR_404_MSG);
+  });
+
+  it('no namespace in pipeline studio page should show 404', async () => {
+    // Go to Pipelines studio
+    await driver.get(`${PIPELINES_URL}/${FAKE_NAMESPACE}/studio`);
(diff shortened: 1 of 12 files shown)


mrahanjam:
You can wait for [this](https://github.com/cdapio/cdap-ui/pull/520) PR to be merged so that you can add the `start_wrangler` command as well to complete your test conversion. (`getArtifactsPoll` is added to utils in that PR)

de-lan:
> You can wait for [this](https://github.com/cdapio/cdap-ui/pull/520) PR to be merged so that you can add the `start_wrangler` command as well to complete your test conversion. (`getArtifactsPoll` is added to utils in that PR)

Yeah, sounds good, will merge after yours

selenium/integration/page.level.errors.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    await driver.findElement(By.css(SELECTOR_404_DEFAULT_MSG));
+  });
+
+  it('no valid path from node server should show 404',async () => {
+    // Go to any random invalid path
+    await driver.get(`${BASE_URL}/randomInvalidPath`);
+    await driver.sleep(500);
+    await driver.findElement(By.css(SELECTOR_404_DEFAULT_MSG));
+  });
+});
```
njbriggs: Nit: newline at EOF

de-lan: updated

selenium/integration/page.level.errors.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+describe('Page level error because of ', () => {
+  beforeAll(async () => {
+    driver = buildChromeDriver();
+    await loginIfRequired(driver);
+  });
+
+  it('no namespace in home page should show 404', async () => {
+    // Go to home page
+    await driver.get(`${CDAP_URL}/${FAKE_NAMESPACE}`);
+    await driver.sleep(1000);
```
seanfeld: @njbriggs @mrahanjam sorry i didn't see this earlier but i think we should figure out a more robust solution rather than `await driver.sleep(1000);` - this introduces flakiness as well as making each test potentially take longer than it should.

```
let ele = await driver.wait(until.elementLocated(By.css(SELECTOR_404_MSG)),10000);
let foo = await ele.getText();
assert(foo == "Whatever is on the 404 msg");
```

https://www.selenium.dev/documentation/webdriver/waits/ -- thats the way that selenium recommends we do this.

The difference is that retries finding the element for 10000ms vs arbitrarily waiting for 1000ms . We can make helper functions to reduce the boilerplate.

mrahanjam: I agree that we could reduce the falkiness but using waits. we have an example [here](https://github.com/cdapio/cdap-ui/blob/a2ce302a55b62a4ba6054b8a32cb187e6da65876/selenium/integration/admin.spec.ts#L59) with the required constants set.

mrahanjam: @de-lan You could add the helper like below to the utils.ts file
```javascript
export const waitUntilElementLocated = async (driver, cssLocator) => {
  const element = await driver.wait(until.elementLocated(By.css(cssLocator)), TEST_TIMEOUT_TIME, TEST_TIMEOUT_MESSAGE, RETRY_INTERVAL);
  return element;
}
```
and then use it like 
```javascript
const keyEl = await waitUntilElementLocated(driver, addedKeyCssLocator);
```

mrahanjam: on second thought, this will help with locating elements only (not navigating through urls). feel free to make it more generic or create a separate one for `driver.get` ones. this example from selenium docs might help:
```javascript
const documentInitialised = () =>
    driver.executeScript('return initialised');

await driver.get('file:///race_condition.html');
await driver.wait(() => documentInitialised(), 10000);
const element = driver.findElement(By.css('p'));
assert.strictEqual(await element.getText(), 'Hello from JavaScript!');
```

de-lan: @mrahanjam Thanks! Yeah I'll add both and update related tests

de-lan:
Dropping this one since we're using Cucumber testing

PR #480, opened 2022-05-02
CDAP-19158 Upload schema when parsing from source

njbriggs:
# CDAP-19158 Upload schema when parsing from source

## Description
Adds the option to upload a schema when parsing at the source. This is necessary for some types (e.g. avro) and optional for others. Will be cherry-picked to 6.7.

## PR Type
- [ ] Bug Fix
- [X] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19158](https://cdap.atlassian.net/browse/CDAP-19158)

## Test Plan
Manually verify

## Screenshots
image

Information tooltip:
image

After importing schema:
image

PR #522, opened 2022-05-25
chore: migrate the file select component in toppanel

de-lan:
## Description
React migration/convert select file

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Test Plan
Manual

### DIFF
--- a/app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
+++ b/app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
@@ -16,12 +16,14 @@
 
 import IconSVG from 'components/shared/IconSVG';
 import React from 'react';
+import { objectQuery } from 'services/helpers';
 import {
   ActionButtonsContainer,
   BorderRightButton,
   ButtonLabel,
   CommonButton,
   CustomTooltip,
+  HiddenInput,
   IconLoading,
   IconPlay,
   IconSchedule,
@@ -49,6 +51,7 @@ interface IActionButtonsProps {
   onSaveDraft: () => void;
   onPublish: () => void;
   onImport: () => void;
+  onFileSelect: (files: FileList) => void;
   onExport: () => void;
   onClickLogs: () => void;
   previewLoading: boolean;
@@ -75,6 +78,7 @@ export const ActionButtons = ({
   onSaveDraft,
   onPublish,
   onImport,
+  onFileSelect,
   onExport,
   onClickLogs,
   previewLoading,
@@ -87,6 +91,16 @@ export const ActionButtons = ({
   viewLogs,
   timerLabel,
 }: IActionButtonsProps) => {
+  const handleFile = (event) => {
+    if (!objectQuery(event, 'target', 'files', 0)) {
+      return;
+    }
+    const uploadedFiles = event.target.files;
+
+    if (onFileSelect) {
+      onFileSelect(uploadedFiles);
+    }
+  };
   return (
     <ActionButtonsContainer>
       {previewMode ? (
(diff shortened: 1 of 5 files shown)


app/hydrator/templates/create/toppanel.html
```diff
@@ -294,6 +295,13 @@
         </div>
       </div>
     </div>
+    <my-file-select
```
de-lan, line 298: Original code moved up into the action buttons wrap div

app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
```diff
@@ -252,6 +267,17 @@ export const ActionButtons = ({
               <ButtonLabel>Import</ButtonLabel>
             </div>
           </CommonButton>
+          <Label className="import-file-label" for="pipeline-import-config-link">
```
njbriggs: Not for this PR, but we should create a file upload component. I started separating logic in #480 but it needs more work and generalization.

app/cdap/components/hydrator/components/TopPanel/ActionButtons/styles.ts
```diff
@@ -18,6 +18,7 @@ import styled from 'styled-components';
 import { Tooltip, withStyles } from '@material-ui/core';
 import IconSVG from 'components/shared/IconSVG';
 import PrimaryTextButton from 'components/shared/Buttons/PrimaryTextButton';
+import { Input } from 'reactstrap';
```
njbriggs: Do we need to use `reactstrap`? Can we just use a plain `input`?

app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
de-lan: Yeah I found that one and tried using it. However this toppanel case is weird it needs two onClick handler because of the save modal check requirement inside the angular controller. We'll need to fox that when we do state migration in the end

app/cdap/components/hydrator/components/TopPanel/ActionButtons/styles.ts
de-lan: Seems no need, removed

app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
```diff
@@ -252,6 +266,17 @@ export const ActionButtons = ({
               <ButtonLabel>Import</ButtonLabel>
             </div>
           </CommonButton>
+          <label htmlFor="pipeline-import-config-link">
+            {/* The onClick here is to clear the file, so if the user uploads the same file
+            twice then we can show the error, instead of showing nothing */}
+            <HiddenInput
```
seanfeld, line 272: so this hidden input was necessary?

de-lan: yes, basically it plays the same functionality as the one in angular

app/cdap/components/hydrator/components/TopPanel/ActionButtons/styles.ts
```diff
@@ -106,3 +107,7 @@ export const RunTimeSpan = styled.span`
   font-weight: 500;
   vertical-align: middle;
 `;
+
+export const HiddenInput = styled.input`
+  display: none !important;
+`; // beating specificity. The file input needs to be hidden at all times
```
seanfeld, line 114: @njbriggs are we okay with using `!important`? 

another way of doing this is 
```
&&& {
    display: none;
}
```

which seems to be the recommended way of doing this in styled components https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity

seanfeld (commented):
I don't think we should be using important - it's usually considered a bad practice.

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception

app/cdap/components/hydrator/components/TopPanel/ActionButtons/styles.ts
de-lan: great point! Updated

seanfeld (approved):

app/cdap/components/hydrator/components/TopPanel/ActionButtons/styles.ts
njbriggs: Sounds like we should start using this

PR #530, opened 2022-05-26
remove matrix from the workflow file

mrahanjam:
# Remove version matrix from the workflow

## Description
Looks like having the version matrix in our workflow file conflicts with `haya14busa/action-workflow_run-status` action that we need to keep our PRs in sync with the status of the github action.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [X] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@
 # This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-name: build
+name: Build with Unit and Integration Tests
 
 on:
   workflow_run:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -50,22 +45,22 @@ jobs:
     - name: Install Ubuntu debendency
       run: sudo apt-get install -y libgif-dev
 
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js 10.16
```
de-lan, line 48: Does this mean that we'll need to manually update these versions number in future?

sumengwang (approved):

.github/workflows/build.yml
mrahanjam: yes using matrix makes the step on line 38 fail. It's a limit imposed by that helper action.

de-lan (approved):

PR #539, opened 2022-06-10
improve configure form anchor el

sumengwang:
# improve configure form anchor el

### DIFF
--- a/app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
+++ b/app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
@@ -15,7 +15,7 @@
  */
 
 import IconSVG from 'components/shared/IconSVG';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { objectQuery } from 'services/helpers';
 import PipelineConfigurationsStore, {
   ACTIONS as PipelineConfigurationsActions,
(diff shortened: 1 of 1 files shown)


app/cdap/components/hydrator/components/TopPanel/ActionButtons/index.tsx
```diff
@@ -133,7 +133,9 @@ export const ActionButtons = ({
     }
   };
 
-  const configureButtonRef = useRef(null);
+  const configureButtonAnchorEl = previewMode
```
seanfeld: the anchor element is just a string? if this is the how you want to do it we should probably rename this variable since its no longer a reference to an anchor el but a selector or something?

sumengwang: ok. I saw `scheduler` is doing this way so I copied. https://github.com/cdapio/cdap-ui/blob/react-migration/toppanel_improvement/app/hydrator/templates/create/toppanel.html#L351

seanfeld: Yeah I think its probably fine as the selector - I was more saying change the variable name to include selector or something like `anchorElSelector` or something like that - I think this is fine either way though because that component specifically allows both string and element.

seanfeld (approved):
No need to rename that variable

de-lan (approved):

PR #545, opened 2022-06-29
Angular to React migration - left panel store migration

seanfeld:
# Left Panel Store Migration

## Description
Lots of changes - left panel store usable through react and angular
new left panel store uses redux toolkit
left the store being called from inside the angular controller, otherwise it gets a little bit too tricky having the separation

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [18832](https://cdap.atlassian.net/browse/CDAP-18832)


njbriggs (approved):

mrahanjam (approved):

de-lan:
Looks like the background of tooltip is not right?
image

app/cdap/components/hydrator/components/TopPanel/TopPanel.tsx
```diff
@@ -62,14 +62,8 @@ const TopPanelContainer = styled.div`
   border-left: 0;
   border-top: 0;
   display: flex;
-  position: fixed;
-  left: 290px;
```
de-lan, line 66: Thanks for fixing this!

seanfeld:
> Looks like the background of tooltip is not right? image

I don't understand why github doesn't let me reply to your comment directly. This is stupid.

Anyways, thats new. I'll fix it now.

app/cdap/components/hydrator/helpers/DAGhelpers.tsx
```diff
@@ -0,0 +1,404 @@
...
+    const location = graphNodes[node.name];
+    let locationX = location.x;
+    let locationY = location.y;
+
+    if (node.type === 'alertpublisher' || node.type === 'errortransform') {
+      const connToThisNode = connections.find((conn) => conn.to === node.name);
+      if (connToThisNode) {
+        const sourceNode = connToThisNode.from;
+        let nonErrorsAlertsConnCount = 0;
+        for (let i = 0; i < connections.length; i++) {
```
de-lan, line 320: node is giving me this: `prefer-for-of: Expected a 'for-of' loop instead of a 'for' loop with this simple iteration`

de-lan (approved):
just some nits

app/cdap/components/hydrator/helpers/DAGhelpers.tsx
seanfeld: i see that too - i don't know where that rule is coming from but i disagree with it. There is nothing wrong with using a plain for loop... will change the tslint

PR #548, opened 2022-07-01
Add 'Average' Directive under 'Calculate' in Wrangler 

adinilfeld:
# Add 'Average' Directive under 'Calculate' in Wrangler 

## Description
Added support for arithmetic operations which take >3 columns (i.e. "Multicolumn" operations), as well as ones which can take columns of different types simultaneously as input. In particular, added a new arithmetic operation, "Average", which is housed under Calculate and averages any number of columns, with any numerical types (besides BigDecimal or Byte).

Note that these changes rely on changes in `data-integrations/wrangler`; see [PR #568](https://github.com/data-integrations/wrangler/pull/568).

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Links
Jira: [PLUGIN-641](https://cdap.atlassian.net/browse/PLUGIN-641)

## Screenshots

Screenshot 2022-07-01 1 54 44 PM

### DIFF
--- a/app/cdap/components/DataPrep/Directives/Calculate/index.js
+++ b/app/cdap/components/DataPrep/Directives/Calculate/index.js
@@ -44,19 +44,40 @@ export default class Calculate extends Component {
     super(props);
     this.VALID_TYPES = NUMBER_TYPES.concat(['string']);
 
-    const {isDisabled, crossColumn, columns, columnType} = this.parseColumns();
+    const {
+      CALCULATE_OPTIONS,
+      CROSS_COLUMN_CALCULATE_OPTIONS,
+      MULTI_COLUMN_CALCULATE_OPTIONS
+    } = this.getCalculateOptions();
+
+    this.CALCULATE_OPTIONS = CALCULATE_OPTIONS;
+    this.CROSS_COLUMN_CALCULATE_OPTIONS = CROSS_COLUMN_CALCULATE_OPTIONS;
+    this.MULTI_COLUMN_CALCULATE_OPTIONS = MULTI_COLUMN_CALCULATE_OPTIONS;
+
+    const {
+      isDisabled,
+      crossColumn,
+      multiColumn,
+      columns,
+      columnTypes,
+      areColsMixedTypes,
+      availableOptions
+    } = this.parseColumns();
 
     this.isDisabled = isDisabled;
     this.crossColumn = crossColumn;
+    this.multiColumn = multiColumn;
     this.columns = columns;
-    this.columnType = columnType;
+    this.columnTypes = columnTypes;
+    this.areColsMixedTypes = areColsMixedTypes;
+    this.availableOptions = availableOptions;
 
     this.defaultState = {
       operationPopoverOpen: null,
       operationInput: 1,
       createNewColumn: false,
       newColumnInput: this.props.column + T.translate(`${COPY_NEW_COLUMN_PREFIX}.inputSuffix`),
-      isDisabled: this.VALID_TYPES.indexOf(this.columnType) === -1 || this.isDisabled,
+      isDisabled: !this.columnTypes.every(type => this.VALID_TYPES.includes(type)) || this.isDisabled
     };
 
     this.state = Object.assign({}, this.defaultState);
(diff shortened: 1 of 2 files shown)


app/cdap/components/DataPrep/Directives/Calculate/index.js
```diff
@@ -44,19 +44,22 @@ export default class Calculate extends Component {
     super(props);
     this.VALID_TYPES = NUMBER_TYPES.concat(['string']);
 
-    const {isDisabled, crossColumn, columns, columnType} = this.parseColumns();
+    const {isDisabled, crossColumn, multiColumn, columns, columnTypes, areColsMixedTypes} = this.parseColumns();
```
njbriggs: With this many fields, it's best to split them up on separate lines. So this should look like
```
    const {
      isDisabled,
      crossColumn,
      multiColumn,
      columns,
      columnTypes,
      areColsMixedTypes
    } = this.parseColumns();
```

app/cdap/components/DataPrep/Directives/Calculate/index.js
```diff
@@ -44,19 +44,22 @@ export default class Calculate extends Component {
...
+    this.columnTypes = columnTypes;
+    this.areColsMixedTypes = areColsMixedTypes;
 
     this.defaultState = {
       operationPopoverOpen: null,
       operationInput: 1,
       createNewColumn: false,
       newColumnInput: this.props.column + T.translate(`${COPY_NEW_COLUMN_PREFIX}.inputSuffix`),
-      isDisabled: this.VALID_TYPES.indexOf(this.columnType) === -1 || this.isDisabled,
+      isDisabled: !this.columnTypes.every(type => this.VALID_TYPES.includes(type)) // Check all columnTypes valid
```
njbriggs: I don't think the comment is necessary - if the function of the code can be easily inferred, you don't need to comment. If you do want to keep the comment, make sure it doesn't split up the expression.

app/cdap/components/DataPrep/Directives/Calculate/index.js
```diff
@@ -466,27 +491,37 @@ export default class Calculate extends Component {
...
+        columns = columns.filter(col => col !== this.props.ddSelected);
+        columns.unshift(this.props.ddSelected);
+
         return {
-          isDisabled: false,
-          crossColumn: true,
+          /*
+          Disable Calculate if both numerical and String-type columns are selected.
+          This could be changed in the future if some options can operate on both at once.
+          */
```
njbriggs: Can this be detected by the type restrictions of the directives? I.e. if no directives are available, disable the menu?

njbriggs (commented):
Overall this PR looks good. I have a couple formatting suggestions and 1 question about how we decide to disable the menu.

app/cdap/components/DataPrep/Directives/Calculate/index.js
adinilfeld: I thought this wasn't possible at first, but it ended up just needing a refactoring of `parseColumns`. It looks a lot different now, let me know if you prefer it this way or how it was before.

app/cdap/components/DataPrep/Directives/Calculate/index.js
adinilfeld: Good point -- removed!

app/cdap/components/DataPrep/Directives/Calculate/index.js
adinilfeld: Done!

app/cdap/components/DataPrep/Directives/Calculate/index.js
```diff
@@ -466,51 +511,47 @@ export default class Calculate extends Component {
...
-      let index = columns.indexOf(this.props.ddSelected);
-      if (type1 === type2 && index !== -1) {
-        columns = index === 0 ? columns : [this.props.ddSelected, columns[0]];
-        return {
-          isDisabled: false,
-          crossColumn: true,
-          columns: columns,
-          columnType: type1
-        };
+    let columnTypes = [...new Set(columns.map(col => DataPrepStore.getState().dataprep.typesCheck[col]))];
```
seanfeld: do these all get reassigned? I'm not looking at the whole file but only use let if its _possibly_ getting reassigned

app/cdap/components/DataPrep/Directives/Calculate/index.js
```diff
@@ -466,51 +511,47 @@ export default class Calculate extends Component {
...
-    type1 = DataPrepStore.getState().dataprep.typesCheck[columns];
     return {
-      isDisabled: true,
-      crossColumn: false,
-      columns: columns,
-      columnType: type1
+      isDisabled: isDisabled,
+      crossColumn: crossColumn,
+      multiColumn: multiColumn,
+      columns: multiColumn ? columns[0] : columns,
```
seanfeld: if multicolumn use only the first column?

seanfeld (approved):
just nits

app/cdap/components/DataPrep/Directives/Calculate/index.js
adinilfeld: Not sure why I had this -- removed, thanks!

app/cdap/components/DataPrep/Directives/Calculate/index.js
adinilfeld: These don't get reassigned later in the code, but I was under the impression that let should be preferred -- if that's not the case, I can change this

njbriggs: `const` is preferred, unless it needs to be `let`

adinilfeld: Ok, makes sense! I've changed it in the code I added.

njbriggs (approved):
LGTM! Just squash before merging

PR #561, opened 2022-07-14
[CDAP-17619]Fix for REST imported pipelines are not exported

radhikav1:
# [CDAP-17619](https://cdap.atlassian.net/browse/CDAP-17619)Allow pipelines exported from REST api to be imported to UI.

## Description
Pipelines which are exported through REST api comes with 'configuration' to hold  pipeline's payload. 
So, parsing the 'configuration' node while importing  the pipelines which are exported through REST api.

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
 [CDAP-17619](https://cdap.atlassian.net/browse/CDAP-17619)

## Test Plan
Tested all the scenarios manually.

### DIFF
--- a/app/cdap/components/PipelineList/EmptyList/index.tsx
+++ b/app/cdap/components/PipelineList/EmptyList/index.tsx
@@ -17,7 +17,7 @@
 import * as React from 'react';
 import { getCurrentNamespace } from 'services/NamespaceStore';
 import { GLOBALS } from 'services/global-constants';
-import { validateImportJSON } from 'services/PipelineErrorFactory';
+import { validateImportJSON, adjustConfigNode } from 'services/PipelineErrorFactory';
 import { Input, Label } from 'reactstrap';
 import uuidV4 from 'uuid/v4';
 import { objectQuery } from 'services/helpers';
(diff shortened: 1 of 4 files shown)


app/hydrator/services/hydrator-upgrade-service.js
```diff
@@ -260,7 +260,7 @@ class HydratorUpgradeService {
 
     reader.onload =  (evt) => {
       let fileDataString = evt.target.result;
-      let isNotValid = this.NonStorePipelineErrorFactory.validateImportJSON(fileDataString);
+      let isNotValid = this.NonStorePipelineErrorFactory.validateImportJSON(fileDataString);
```
sumengwang: no need the white space

app/cdap/components/PipelineList/EmptyList/index.tsx
```diff
@@ -89,13 +89,13 @@ class EmptyList extends React.PureComponent<IProps, IState> {
         return;
       }
 
-      const fileDataString: string = evt.target.result;
+      let fileDataString: string = evt.target.result;
```
sumengwang, line 92: Can you check if the error check requires the original `fileDataString`, if it doesn't care you can just combine the two lines into one

radhikav1: Once original `fileDataString` is validated and doesn't have any errors, then only it is required to adjust the config node. So, both are not combined here.

app/hydrator/services/hydrator-upgrade-service.js
njbriggs: +1 - please remove trailing whitespace

njbriggs (approved):
I think the trailing whitespace can be cleaned up - after that, LGTM

de-lan (approved):

sumengwang (approved):

PR #568, opened 2022-07-20
CDAP-19297: Automatically delete rejected request after rejection in cdf

mrahanjam:
# TITLE

## Description
This PR automatically deletes the rejected requests on cdf side to enable hdf to retry the connection upon rejection.

## PR Type
- [ ] Bug Fix
- [X] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19297](https://cdap.atlassian.net/browse/CDAP-19297)

## Test Plan
Manual

## Screenshots
N/A

### DIFF
--- a/app/cdap/components/Administration/TetheringTabContent/reducer.ts
+++ b/app/cdap/components/Administration/TetheringTabContent/reducer.ts
@@ -36,6 +36,11 @@ export const initialConnectionsState = {
   establishedConnections: [],
 };
 
+const apiConnectionActions = {
+  ACCEPT: 'accept',
+  REJECT: 'reject',
+};
+
 enum IConnectionsActions {
   SET_CONNECTIONS,
   DELETE_CONNECTION,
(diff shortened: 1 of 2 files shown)


sumengwang (approved):

app/cdap/components/Administration/TetheringTabContent/reducer.ts
```diff
@@ -36,6 +36,11 @@ export const initialConnectionsState = {
   establishedConnections: [],
 };
 
+const apiConnectionActions = {
```
njbriggs, line 39: Can this be defined in one place?

mrahanjam: I wanted to do that initially, but we don't keep constants at an appropriate place for these two files to access. There is one `constants.ts` file for tethering table, and one for new requests. Didn't make sense to put this in either. Also didn't make sense to create a constant file just for this one.

njbriggs (approved):

PR #572, opened 2022-07-21
fix: propagate stateParams correctly

de-lan:
## Description
The pipelineTriggersSidebars is gone after leftPanel react migration. Along with minor fix

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19436](https://cdap.atlassian.net/browse/CDAP-19436)


sumengwang (approved):

njbriggs (approved):

mrahanjam (approved):
LGTM, the only questionable part in the old code is the `ng-if="false"` on line 33. Not sure why it's there to begin with.

de-lan:
> LGTM, the only questionable part in the old code is the `ng-if="false"` on line 33. Not sure why it's there to begin with.

That should be added with the migration, disabling the old cold and render the react code

PR #574, opened 2022-07-26
Testing/cdap 19219

mrahanjam:
# Convert admin.spec.ts to selenium

## Description
This PR converts `admin.spec.ts` to use the E2E test framework.

** UPDATE TO THE STEPS BELOW: You can skip the first three steps and start at step 4. In case of failure, run the first three steps.

Steps to run the tests locally:

1. Clone the `cdap-e2e-tests` repo from [here](https://github.com/cdapio/cdap-e2e-tests)
2. `cd` into the `cdap-e2e-tests` directory and run the following command: `mvn clean install`
3. Make sure `cdap sandbox` and `cdap-ui server` are running
4. `cd` into `cdap-ui` directory
5. Make sure you are on `testing/CDAP/19219` branch
6. Try running the following command: `mvn clean verify -P e2e-tests`

In case of failure, you most likely need to do the following if it's your first time running tests using `cdap-e2e-tests`:
- open `cdap-ui` project with `intelliJ`
- click on the Maven taskbar on the right side of the window
- expand `profiles` and make sure `e2e-test` is checked. Then click on `Reload all Maven Projects`
Screen Shot 2022-07-28 at 9 26 49 AM

- right click on the `TestRunner` class in `src/e2e-test/java/io.cdap.cdap.ui/runners/TestRunner`, and click on `Run 'Test Runner'`
Screen Shot 2022-07-28 at 9 30 36 AM

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [X] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19219](https://cdap.atlassian.net/browse/CDAP-19219)

## Test Plan
Automated

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,19 @@
       Only update this when releasing a new UI pack.
     -->
     <ui.pack.version>4.3.4_p7</ui.pack.version>
+    <guava.version>27.0.1-jre</guava.version>
+    <testSourceLocation>${project.basedir}/src/test/java/</testSourceLocation>
   </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>${guava.version}</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
 
   <dependencies>
     <dependency>
@@ -68,6 +80,7 @@
   </dependencies>
 
   <build>
+    <testSourceDirectory>${testSourceLocation}</testSourceDirectory>
     <plugins>
       <plugin>
         <artifactId>maven-clean-plugin</artifactId>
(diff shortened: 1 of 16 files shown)


pom.xml
```diff
@@ -36,8 +36,20 @@
       Only update this when releasing a new UI pack.
     -->
     <ui.pack.version>4.3.4_p7</ui.pack.version>
+    <guava.version>27.0.1-jre</guava.version>
```
itsankit-google, line 39: I am not sure if it is right to change guava version like that, 
cdap uses different guava version - https://github.com/cdapio/cdap/blob/develop/pom.xml#L134

@tivv wdyt?
@jockech did some workaround to use suitable guava version required for e2e-tests in this commit - https://github.com/data-integrations/amazon-s3-plugins/commit/d176ef9a8149791f652f4ea2b7a85c61ceeef0e0

mrahanjam: This workaround was suggested by @tivv as Selenium is incompatible with guava 13.0.1. ~I'm thinking with the removal of `cdap` as parent in my follow-up PR (will be merged after this one is merged), I can remove guava from this `pom.xml` file altogether.~

itsankit-google: can we add it only in the `e2e-tests` profile?

mrahanjam: yes, we most probably can. I'll add it to my other PR though as that's where these tests will start running in automation.

itsankit-google: If we are removing cdap as parent in follow up PR, removing guava altogether from pom sounds good.
Then we don't even need it in `e2e-tests` profile.

mrahanjam: Actually I take that back, UI uses guava here:
https://github.com/cdapio/cdap-ui/blob/3c7f843ab4bc19585999e0adeef3a2de0058349a/src/main/java/io/cdap/cdap/ui/ConfigurationJsonTool.java#L19

itsankit-google: Ahh I see, then I would suggest doing something similar what @jockech did. Adding `v13.0.1` as direct dependency and override it in `e2e-tests` profile with `27.0.1-jre`.

mrahanjam: Yup, trying that right now

mrahanjam: @itsankit-google on second thought, why would it hurt to set the guava version to 27? The only change with that happens here and UI uses guava very limitedly
https://github.com/cdapio/cdap-ui/blob/3c7f843ab4bc19585999e0adeef3a2de0058349a/src/main/java/io/cdap/cdap/ui/ConfigurationJsonTool.java#L72

jockech: > Ahh I see, then I would suggest doing something similar what @jockech did. Adding `v13.0.1` as direct dependency and override it in `e2e-tests` profile with `27.0.1-jre`.

yes. it seems the `profile` can only override a explicitly declared dependency in the POM.

tivv: You should be able to use dependency management in the profile.

mrahanjam: jockech@ approach doesn't work in the UI repo, running the `mvn clean verify -P e2e-tests` errors out with
Screen Shot 2022-07-29 at 8 41 53 PM
The only way I've managed to make it work is what I currently have in this PR which Vitalli and I came up with. I'd like to know what's the problem considering there'll be no dependency on cdap eventually which uses a different guava version.

src/e2e-test/features/admin.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  @ADMIN_TEST
+  Scenario: Allow user to save valid preference at instance level after fixing error
+    When Open configuration page
+    Then Click on "System Preferences" accordion
+    Then Click on "Edit System Preferences" button
+    Then Add "name" as key
+    Then Add "hello" as value
+    Then Click on "Save Preferences" button
+    Then Verify successful saving of preferences with "name" as key and "hello" as value
```
njbriggs: Nit: newline at EOF

src/e2e-test/java/io/cdap/cdap/ui/stepsdesign/Admin.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Then("Verify failure in saving changes")
+  public void checkForErrorInSavingPreferences() {
+    ElementHelper.clickOnElement(
+      Helper.locateElementByCssSelector("div[class=\"preferences-error\"]")
+    );
+  }
+
+  @Then("Verify successful saving of preferences with {string} as key and {string} as value")
```
njbriggs, line 85: Arguments are mapped by position?

tivv: Usually yes in java as argument names are not available in reflection by default (they are part of debug information only, method signature only have list of arguments with types but without names)

njbriggs (approved):

de-lan (approved):

PR #582, opened 2022-08-04
Add HTTP request handler to e2e tests

mrahanjam:
# Add HTTP request handler to e2e tests

## Description
This PR adds a HTTP request handler class for the e2e tests to use.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [X] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
N/A

## Test Plan
N/A

## Screenshots
N/A

### DIFF
--- a/src/e2e-test/java/io/cdap/cdap/ui/utils/HttpRequestHandler.java
+++ b/src/e2e-test/java/io/cdap/cdap/ui/utils/HttpRequestHandler.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.cdap.ui.utils;
+
+import com.google.common.base.Joiner;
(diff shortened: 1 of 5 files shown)


src/e2e-test/java/io/cdap/cdap/ui/utils/HttpRequestHandler.java
```diff
@@ -0,0 +1,73 @@
...
+    String bodyStr = Joiner.on(",").withKeyValueSeparator("=").join(body);
+    return new ContentProvider<InputStream>() {
+      @Override
+      public InputStream getInput() {
+        return new ByteArrayInputStream(bodyStr.getBytes());
+      }
+    };
+  }
+
+  private static Multimap<String, String> castMapToMultiMap(Map<String, String> header) {
```
mrahanjam, line 66: Added this to keep the type of request body and headers consistent. I think it avoids confusion at a higher level (see usage in `loginIfRequired` method of helper class)
https://github.com/cdapio/cdap-ui/blob/6e1f764232a9b8d1890638d58d909265a62893b1/src/e2e-test/java/io/cdap/cdap/ui/utils/Helper.java#L54

src/e2e-test/features/admin.feature
```diff
@@ -30,6 +31,7 @@ Feature: Admin - Validate profile setting
 
   @ADMIN_TEST
   Scenario: Allow user to save valid preference at instance level after fixing error
+    Given Login if required
```
njbriggs: Is there an equivalent of `before` or `beforeEach` in Cucumber?

njbriggs (approved):
One question that might help make cleaner tests, otherwise LGTM

src/e2e-test/features/admin.feature
mrahanjam: I think there is but I am not sure if it's for all feature files or can be tied to a single feature file (we don't call the `loginIfRequired` method in all feature files). I'll update once I have a better idea

sumengwang (approved):

de-lan (approved):
LGTM

PR #592, opened 2022-08-12
Pom/node version

seanfeld:
# Fixit update node version 

## Description
Update node version because github actions build allows us to move off bamboo and node 10

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,7 +251,7 @@
                   <goal>install-node-and-yarn</goal>
                 </goals>
                 <configuration>
-                  <nodeVersion>v10.16.2</nodeVersion>
+                  <nodeVersion>v16.16.0</nodeVersion>
                   <yarnVersion>v1.7.0</yarnVersion>
                 </configuration>
               </execution>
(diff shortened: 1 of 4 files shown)


mrahanjam:
We probably wanna update node version in the workflow as well:
https://github.com/cdapio/cdap-ui/blob/6d349e236c3d58b77b660168508597a2b7e35605/.github/workflows/build.yml#L44

mrahanjam (approved):
LGTM, left a note on updating github workflow

seanfeld:
> We probably wanna update node version in the workflow as well:
> 
> https://github.com/cdapio/cdap-ui/blob/6d349e236c3d58b77b660168508597a2b7e35605/.github/workflows/build.yml#L44

good point I wasn't very aware of that file.

pom.xml
```diff
@@ -368,7 +368,7 @@
                   <goal>install-node-and-yarn</goal>
                 </goals>
                 <configuration>
-                  <nodeVersion>v10.16.2</nodeVersion>
+                  <nodeVersion>v16.16.0</nodeVersion>
                   <yarnVersion>v1.7.0</yarnVersion>
```
mrahanjam, line 372: Don't we wanna update `yarn` as well?

seanfeld: no that should be a separate pr

itsankit-google (approved):

sumengwang:
Looks like some e2e tests are failing. Is there any concern?

PR #596, opened 2022-08-17
fixed existing cucumber tests

sumengwang:
Fixed broken cucumber tests after merging react migration

### DIFF
--- a/app/cdap/components/hydrator/components/SidePanel/SidePanel.tsx
+++ b/app/cdap/components/hydrator/components/SidePanel/SidePanel.tsx
@@ -202,7 +202,7 @@ export const SidePanel = ({
 
   const renderPlugins = (plugins: [any]) => {
     return plugins.map((plugin) => {
-      const id = `plugin-${plugin.displayName}-${plugin.type}`;
+      const id = `plugin-${plugin.name}-${plugin.type}`;
       const label = plugin.displayName || plugin.name;
       return (
         <Tooltip
@@ -225,6 +225,7 @@ export const SidePanel = ({
             key={id}
             onClick={(event) => handleClickPlugin(event, plugin)}
             data-cy={id}
+            data-testid={id}
           >
             {sidePanelViewType === 'icon' && (
               <>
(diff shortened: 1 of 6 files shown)


src/e2e-test/features/pipeline.hierarchy.feature
```diff
@@ -21,6 +21,7 @@ Feature: Pipeline Hierarchy - Hierarchy Widgets
   Scenario: Render file and duplicate record
     Given Open Datafusion Project to configure pipeline
     Then Add "File" node to canvas
+    Then Close "Source" panel
```
sumengwang, line 24: I need to do this because the side panel accordian is expanding too long. Do you know how to make it fixed height and overflow-y, like the original one? @seanfeld

app/cdap/components/hydrator/components/SidePanel/SidePanel.tsx
```diff
@@ -271,12 +272,12 @@ export const SidePanel = ({
           square
           expanded={expanded}
           onClick={() => handleOpenAccordion(group.name)}
+          data-cy={`plugin-${group.name}-group`}
```
mrahanjam, line 275: missing `data-testid`?

src/e2e-test/java/io/cdap/cdap/ui/utils/Helper.java
```diff
@@ -113,16 +117,16 @@ public static void deployAndTestPipeline(String filename, String pipelineName) {
...
     WaitHelper.waitForPageToLoad();
     WebElement uploadFile = SeleniumDriver.getDriver()
-            .findElement(By.xpath("//*[@id='pipeline-import-config-link']/input[@type='file']"));
+            .findElement(By.xpath("//*[@id='pipeline-import-config-link']"));
 
     File pipelineJSONFile = new File(Constants.FIXTURES_DIR + filename);
     String filePath = pipelineJSONFile.getAbsolutePath();
     uploadFile.sendKeys(filePath);
 
-    String pipelineNameCSSSelector = "div[class*='pipeline-name']";
```
mrahanjam, line 122: why the switch to xpath?

sumengwang: wanted to use xpath contains

de-lan (approved):

src/e2e-test/java/io/cdap/cdap/ui/utils/Helper.java
mrahanjam: Why aren't the old locators working? new classes? I generally prefer css selectors as they're faster than xpath.

sumengwang: yep new classes

mrahanjam (approved):

src/e2e-test/features/pipeline.hierarchy.feature
seanfeld: What do you mean it's expanding too long? its expanding passed the boundaries? if its covering the footer area scrolling that means something was introduced that made it do that.

The old version had it so that only one accordion could be open at once - which is a behavior we don't want - so if that is what you're seeing that is expected behavior.

PR #600, opened 2022-08-19
fix metadata url

sumengwang:
update the metadata url and fixed cypress test


mrahanjam (approved):

de-lan (approved):

de-lan:
Do we know why it was working before and now need to update the url?

PR #615, opened 2022-08-31
fix flaky trigger test

sumengwang:


cypress/integration/pipeline.triggers.spec.ts
```diff
@@ -77,6 +77,7 @@ describe('Pipeline Studio', () => {
         cy.get(dataCy(`${TRIGGER_PIPELINE_1}-expanded`)).get(dataCy('disable-trigger-btn')).click();
         cy.get(dataCy('Delete')).should('exist');
         cy.get(dataCy('Delete')).click();
+        cy.get(dataCy('enabled-triggers-tab')).click();
```
sumengwang, line 80: @de-lan After deleting the trigger, it will automatically switch to the 'set trigger' tab. I think we want to switch back to the 'enabled trigger' tab to verify the element doesn't exist

seanfeld (approved):

de-lan (approved):

PR #628, opened 2022-09-14
mvn clean on dev-clean profile won't remove node_modules

sumengwang:


sechegaray (approved):

mrahanjam (approved):

tivv (changes requested):
Please use a profile. By default `clean` should clean everything

de-lan (approved):

pom.xml
```diff
@@ -660,5 +658,28 @@
...
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-clean-plugin</artifactId>
+            <version>2.6.1</version>
+            <configuration>
+              <excludeDefaultDirectories>true</excludeDefaultDirectories>
+              <filesets>
+                <fileset>
+                  <directory>node_modules</directory>
+                  <excludes>
+                    <exclude>**</exclude>
```
tivv: Hm Are you sure it deletes anything with this settings? You excluded the defaults and I don't see any includes

sumengwang: You are right. I noticed that the clean without profile will delete the node_modules folder. So In the dev-clean profile, I removed the part that explicitly deletes node_modules

tivv: But are you sure it deletes other stuff? E.g. if target directory is present, does it delete it? Looking at the setting I don't see this settings deleting anything.

sumengwang: Yea I tested it. It will delete target/

tivv (approved):

PR #639, opened 2022-09-23
Fix expected error count

sumengwang:


cypress/integration/plugins.validation.spec.ts
```diff
@@ -93,7 +93,7 @@ describe('Plugin properties', () => {
 
     // Testing for errors found
     cy.get(dataCy(VALIDATE_BTN_SELECTOR)).click();
-    cy.get(dataCy('plugin-properties-errors-found')).should('have.text', '3 errors found.');
+    cy.get(dataCy('plugin-properties-errors-found')).should('have.text', '2 errors found.');
```
radhikav1, line 96: Relaying on number might be always unpredictable I think, instead can't we just check with 'errors found'  string alone?

radhikav1 (approved):

mrahanjam (approved):

cypress/integration/plugins.validation.spec.ts
sumengwang: yea that's what I was thinking too. good point

mrahanjam: I'd argue that in this case number of failures matter. Looks like the tests is based on missing mandatory props. If we only rely on the text, we might miss a missing required prop when validating (e.g., expecting 3 failures while getting 2).

de-lan (approved):

cypress/integration/plugins.validation.spec.ts
sumengwang: so does it make sense if the plugin changes its mandatory fields, we have to maintain the test on the ui side?

mrahanjam: It's not unusual for tests to have to be updated based on new changes.

PR #652, opened 2022-10-03
[CDAP-19653] filter config groups before assign filtered values

sumengwang:
# [CDAP-19653] filter config groups before assign filtered values

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19653](https://cdap.atlassian.net/browse/CDAP-19653)

## Test Plan
manually tested. toggle connection on and off and validate works

[CDAP-19653]: https://cdap.atlassian.net/browse/CDAP-19653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/AbstractWidget/StateWrapper.js
+++ b/app/cdap/components/AbstractWidget/StateWrapper.js
@@ -26,6 +26,7 @@ export default class StateWrapper extends PureComponent {
     onChange: PropTypes.func,
     updateAllProperties: PropTypes.func,
     widgetProps: PropTypes.object,
+    widgetProperty: PropTypes.object,
     extraConfig: PropTypes.object,
     disabled: PropTypes.bool,
     errors: PropTypes.array,
@@ -81,6 +82,7 @@ export default class StateWrapper extends PureComponent {
         onChange={this.onChange}
         updateAllProperties={this.props.updateAllProperties}
         extraConfig={extraConfig}
+        widgetProperty={this.props.widgetProperty}
         disabled={disabled}
         errors={errors}
         dataCy={dataCy}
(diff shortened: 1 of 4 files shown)


app/cdap/components/shared/ConfigurationGroup/index.tsx
```diff
@@ -161,9 +161,7 @@ const ConfigurationGroupView: React.FC<IConfigurationGroupProps> = ({
 
   function handleValueChanges(changedValues, params: { [key: string]: boolean } = {}) {
     let fcg = filteredConfigurationGroups;
-    if (params.updateFilteredConfigurationGroups) {
-      fcg = filterByCondition(configurationGroups, widgetJson, pluginProperties, changedValues);
-    }
+    fcg = filterByCondition(configurationGroups, widgetJson, pluginProperties, changedValues);
```
njbriggs: Do we have any cases where `updateFilteredConfigurationGroups` should be false? Or, would it be better to add `updateFilteredConfigurationGroups: true` in the calling function for this scenario?

sumengwang: I found one usage of setting it to true. But I believe this should apply to many components like toggle, radiogroup, select, etc., since they usually change what to show. Then is it really necessary to differentiate the cases? The only downside is one extra function call for all value changes. Let me know what you think

de-lan (approved):

app/cdap/components/shared/ConfigurationGroup/index.tsx
sumengwang: I pushed some new changes that use this flag

seanfeld (approved):

app/cdap/components/shared/ConfigurationGroup/index.tsx
njbriggs: Thanks. I don't remember the exact path, but I think we had concerns about backwards compatibility when we added that functionality.

njbriggs (approved):

PR #767, opened 2022-11-11
[CDAP-19967]: fix view metadata button for plugins with no reference name

mrahanjam:
# Fix view metadata button for plugins with no reference name

## Description
This PR updates the node properties page to fetch the `normalized fqn` in the absence of `reference name` so that navigation to the metadata page is not broken.

## PR Type
- [ ] Bug Fix
- [X] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19967](https://cdap.atlassian.net/browse/CDAP-19967)

## Test Plan
Manual

### DIFF
--- a/app/cdap/components/PipelineDetails/store/ActionCreator.js
+++ b/app/cdap/components/PipelineDetails/store/ActionCreator.js
@@ -296,6 +296,25 @@ const getStatistics = (params) => {
   );
 };
 
+const getMetadataEndpoints = ({ namespace, appId, workflowType, runId }) => {
+  return MyPipelineApi.getMetadataEndpoints({
+    namespace,
+    runId,
+    appId,
+    workflowType,
+  }).subscribe(
+    (res) => {
+      PipelineDetailStore.dispatch({
+        type: ACTIONS.SET_METADATA_ENDPOINTS,
+        payload: { metadataEndpoints: res.endpoints },
+      });
+    },
+    (err) => {
+      console.log(err);
+    }
+  );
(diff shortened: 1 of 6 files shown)


sumengwang (approved):

app/cdap/components/PipelineDetails/store/ActionCreator.js
```diff
@@ -296,6 +296,25 @@ const getStatistics = (params) => {
   );
 };
 
+const getMetadataEndpoints = ({ namespace, appId, workflowType, runId }) => {
```
sumengwang, line 299: should be setMetadataEndpoints?

mrahanjam: I followed the general naming pattern in the `ActionCreator`. I believe the `get` is due to them getting the response from the backend API. `getStatistics`, `getNextRuntime`, etc. do similar stuff.

seanfeld (approved):

app/cdap/components/PipelineDetails/store/ActionCreator.js
sumengwang: ur i see those. I still don't think it quite make sense, because ultimately you are dispatching the SET, even a name like `getThenSet` makes more sense to me. But this is fine as it is.
And you can remove the `return`, since you are not actually using the subscription elsewhere

PR #786, opened 2022-11-16
bugfix/CDAP-20100: add tooltip for composite trigger configuration

de-lan:
Add tooltip for composite trigger configuration

## Description
Summary of changes

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Links
Jira: 
[CDAP-20022](https://cdap.atlassian.net/browse/CDAP-20022)
[CDAP-20100](https://cdap.atlassian.net/browse/CDAP-20100)

## Screenshots
image
image

### DIFF
--- a/app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
+++ b/app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
@@ -46,7 +46,7 @@ import {
   PipelineTriggerHeader,
   SearchTriggerTextField,
 } from 'components/PipelineTriggers/shared.styles';
-import { InputAdornment, TextField } from '@material-ui/core';
+import { InputAdornment, TextField, Tooltip, withStyles } from '@material-ui/core';
 import SearchIcon from '@material-ui/icons/Search';
 import {
   initialAvailablePipelineListState,
@@ -121,6 +121,18 @@ const TriggerConfigHeader = styled.div`
   font-size: 11px;
 `;
 
+const CustomTooltip = withStyles(() => {
+  return {
+    arrow: {
+      color: '#3c4355',
+    },
+    tooltip: {
+      fontSize: '13px',
+      backgroundColor: '#3c4355',
+    },
+  };
+})(Tooltip);
+
 interface IPipelineListCompositeTabViewProps {
   existingTriggers: ISchedule[];
   pipelineList: IPipelineInfo[];
@@ -150,6 +162,8 @@ const PipelineListCompositeTabView = ({
   setTab,
 }: IPipelineListCompositeTabViewProps) => {
   const [state, dispatch] = useReducer(triggerNameReducer, initialAvailablePipelineListState);
+  const emptyTriggerErrorMsg =
+    triggersGroupToAdd.length === 0 ? T.translate(`${PREFIX}.emptyCompositeTriggerError`) : '';
 
   useEffect(() => {
     dispatch({ type: 'SET_NAMESPACE' });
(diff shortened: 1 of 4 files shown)


app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
```diff
@@ -55,6 +55,9 @@ import {
 import PayloadConfigModal from 'components/PipelineTriggers/PayloadConfigModal';
 import PipelineCompositeTriggerRow from './PipelineCompositeTriggerRow';
 
+const colors = require('styles/colors.scss');
+const toolTipBackground = colors.blue07;
```
seanfeld: Can you get this some way that doesn't use colors.scss?

app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
```diff
@@ -338,16 +355,20 @@ const PipelineListCompositeTabView = ({
...
-            color="primary"
-            disabled={state.isNameInvalid || triggersGroupToAdd.length === 0}
-            variant="contained"
-            onClick={() => addGroupTriggerClick()}
-            data-cy="enable-group-trigger-btn"
-            data-testid="enable-group-trigger-btn"
-          >
-            {T.translate(`${PREFIX}.addNewTrigger`)}
-          </EnableGroupTriggerButton>
+          <CustomTooltip arrow title={state.triggerNameError || emptyTriggerErrorMsg}>
```
seanfeld, line 355: Are you sure this should use a tooltip? usually we try not to use tooltips for stuff that that has intractability?

sumengwang: learned a new word today

app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
de-lan: updated, thanks!

app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
de-lan: what's the advice here? I've set the error message in the trigger name input, but the configuration page could be too long if there are many upstream pipelines and users will ignore it. We need a proper way to guide users why the enable button is disabled

sumengwang (approved):

app/cdap/components/PipelineTriggers/PipelineListCompositeTab/index.tsx
seanfeld: use richtooltip

seanfeld: i used it for a similar thing in the sidepanel

seanfeld: its not that a popup menu is bad per se - its using the `tooltip` component to accomplish it.

seanfeld: @sumengwang it should've been https://en.wiktionary.org/wiki/interactability but the dang autocorrect don't like me

seanfeld: nvm i read this wrong - i thought the button was part of the tooltip content 😞

seanfeld (approved):

PR #841, opened 2022-12-08
[CDAP-19232]: convert pipeline context menu to cucumber

mrahanjam:
# Convert pipeline context menu to cucumber

## Description
This PR converts pipeline context menu to cucumber

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [X] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-19232](https://cdap.atlassian.net/browse/CDAP-19232)

## Test Plan
N/A

### DIFF
--- a/src/e2e-test/features/pipeline.contextmenu.feature
+++ b/src/e2e-test/features/pipeline.contextmenu.feature
@@ -1,0 +1,109 @@
+#
+# Copyright © 2022 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+
+@Integration_Tests
+Feature: Pipeline Context Menu
+
+  @PIPELINE_CONTEXT_MENU_TEST
+  Scenario: Show plugin level context menu
+    When Open Pipeline Studio Page
+    When Create a simple pipeline
+    Then Open source node context menu
+    Then Validate Copy Plugin and Delete Plugin options are visible
+
+  @PIPELINE_CONTEXT_MENU_TEST
+  Scenario: Ensure functionality by clicking on plugin context menu options
+    When Open Pipeline Studio Page
+    When Create a simple pipeline
+    Then Delete source node through context menu
+    Then Delete transform node through context menu
+    Then Delete sink node through context menu
+    Then Verify export button is disabled
+    Then Undo the deletions
+    Then Export and verify the pipeline has 2 connections and 3 stages
+
+  @PIPELINE_CONTEXT_MENU_TEST
+  Scenario: Show pipeline level icon
+    When Open Pipeline Studio Page
+    When Create a simple pipeline
+    Then Right click on canvas
+    Then Ensure canvas context menu option "Wrangle" is displayed
+    Then Ensure canvas context menu option "Zoom In" is displayed
+    Then Ensure canvas context menu option "Zoom Out" is displayed
+    Then Ensure canvas context menu option "Fit To Screen" is displayed
+    Then Ensure canvas context menu option "Align" is displayed
+    Then Ensure canvas context menu option "Paste" is displayed
+
+  @PIPELINE_CONTEXT_MENU_TEST
+  Scenario: Ensure clicking zoom in and zoom out works properly
+    When Open Pipeline Studio Page
+    When Create a simple pipeline
+    Then Verify "default" zoom level
+    Then Right click on canvas
+    Then Click on "Zoom In" from canvas context menu option
+    Then Verify "first" zoom level
+    Then Right click on canvas
+    Then Click on "Zoom In" from canvas context menu option
+    Then Verify "second" zoom level
+    Then Right click on canvas
+    Then Click on "Zoom Out" from canvas context menu option
+    Then Verify "first" zoom level
+    Then Right click on canvas
+    Then Click on "Zoom Out" from canvas context menu option
+    Then Verify "default" zoom level
+
+  @PIPELINE_CONTEXT_MENU_TEST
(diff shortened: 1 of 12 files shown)


src/e2e-test/features/pipeline.contextmenu.feature
```diff
@@ -0,0 +1,111 @@
...
+    When Open Pipeline Studio Page
+    When Create a complex pipeline
+    Then Verify complex pipeline nodes are visible
+    Then Zoom in ten times on pipeline canvas
+    Then Verify complex pipeline nodes are not visible
+    Then Right click on canvas
+    Then Click on "Fit To Screen" from canvas context menu option
+    Then Verify complex pipeline nodes are visible
+
+  @PIPELINE_CONTEXT_MENU_TEST @ignore
```
mrahanjam, line 79: The last three scenarios have `@ignore` tags since I haven't figured out a way to give clipboard access to chrome while running in `headless` mode. 
I found a solution on stackoverflow that works well when `headless` mode is disabled, and all these three scenarios passed locally. However, the solution adds a preference to the chrome driver for giving access to clipboard. Preferences are not supported in `headless` mode.
You can read more on the solution and issue here:
[Solution 1](https://stackoverflow.com/questions/53669639/enable-clipboard-in-automated-tests-with-protractor-and-webdriver), [Solution 2](https://testingbot.com/support/selenium/permission-popups#:~:text=Clipboard%20Permissions), [preferences not supported in headless](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1925)

mrahanjam: I'll hold off on merging for another day, but the fix will most likely be on the chrome driver in `e2e-tests` repo. These changes will stay, I'll just remove the @ignore tags once the issue is resolved.

src/e2e-test/java/io/cdap/cdap/ui/stepsdesign/CommonSteps.java
```diff
@@ -56,7 +56,7 @@ public void openPipelineStudioPage() {
       }
     }
     // wait for rendering to finish otherwise elements are not attached to dom
-    Helper.waitSeconds(3);
```
sumengwang, line 59: hmm i see we just keep incrementing the seconds from past commits. seems bad. Maybe we can do something like wait till bq source plugin to display. But it's ok for now

src/e2e-test/java/io/cdap/cdap/ui/utils/Commands.java
```diff
@@ -40,10 +40,21 @@
 
 public class Commands implements CdfHelper {
 
+  // Simple Pipeline Nodes
   public static NodeInfo simpleSourceNode = new NodeInfo("BigQueryTable", "batchsource", "0");
   public static NodeInfo simpleTransformNode = new NodeInfo("Wrangler", "transform", "1");
   public static NodeInfo simpleSinkNode = new NodeInfo("BigQueryMultiTable", "batchsink", "2");
 
+  // Complex Pipeline Nodes
+  public static NodeInfo complexSourceNode1 = new NodeInfo("Spanner", "batchsource", "0");
```
sumengwang, line 49: why are these commands but not constants?

src/e2e-test/java/io/cdap/cdap/ui/stepsdesign/CommonSteps.java
mrahanjam: I agree, tbf I went with 4 in the refactoring PR initially. but if it comes to increasing it once more, will make a diff change something like you suggested.

src/e2e-test/java/io/cdap/cdap/ui/utils/Commands.java
mrahanjam: Good point but I mainly kept them here due to the simple nodes being used across different tests. Would've made this PR messy. Might move them in another refactor PR

src/e2e-test/java/io/cdap/cdap/ui/stepsdesign/PipelineContextMenu.java
```diff
@@ -0,0 +1,211 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class PipelineContextMenu {
+
+  public static final Gson GSON = new Gson();
+  public static String pipelineName = "";
```
sumengwang, line 44: why not do it here `public static String pipelineName = "Test_Pipeline_" + UUID.randomUUID()`;

mrahanjam: Each scenario needs its own UUID (for the sake of independence), declaring at the top means same UUID for all scenarios (class is instantiated only once for the feature file).

src/e2e-test/java/io/cdap/cdap/ui/utils/Commands.java
```diff
@@ -346,4 +339,17 @@ public static int checkDraftPipelineExistsViaApi(String draftId) throws IOExcept
...
+    ElementHelper.clickOnElement(Helper.locateElementByTestId("pipeline-metadata"));
+    ElementHelper.clearElementValue(Helper.locateElementByCssSelector("#pipeline-name-input"));
+    ElementHelper.sendKeys(Helper.locateElementByCssSelector("#pipeline-name-input"), pipelineName);
+    ElementHelper.clickOnElement(
+      Helper.locateElementByTestId("pipeline-metadata-ok-btn"));
+  }
+  
+  public static void dismissTopBanner() {
+    ElementHelper.clickOnElement(Helper.locateElementByXPath(
+      "//div[@data-testid='valium-banner-hydrator']//button[@class='close ng-scope']"));
```
sumengwang, line 353: maybe just add a `valium-banner-hydrator-close-button` datatestid

sumengwang (approved):

seanfeld (approved):

src/e2e-test/java/io/cdap/cdap/ui/utils/Commands.java
mrahanjam: I already added `banner-dismiss-button`, but that only works for failure banners. This was a tricky one to locate on the dom, I believe it has to do with the success banner disappearing after 3 seconds. Only worked with the xpath selector.

PR #842, opened 2022-12-08
Add better typescript and javascript linting/checking etc. Fix a ton of types

seanfeld:
# Add better typescript checking and eslinting

## Description
Fixed a ton of types
Added better eslinting
added google styleguide for typescript checking
changed dev webpack to run just one webpack command

Very much expecting this to break some stuff and for this to not build correctly the first time.

## More Context
The reason there is a huge amount of files in this pr is that i ran eslint --fix with the new rules, which basically ran prettier on basically all of the files. Then after I got the build working again with https://github.com/google/gts the google typescript style guide rules, there was an insane amount of types that were incorrect. I then went through and corrected over a thousand of the incorrect types throughout the application. After that I decided to fix the webpack build a bit so now eslint/typechecking runs after every save and we only use one version of webpack for local development. I'm going to be making more changes there to get things running more smoothly as well. Also webpack reloads the page on a save now.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Links
-- Followup work --
eslint/tsconfig
https://cdap.atlassian.net/browse/CDAP-20186
imports that need to be changed/updated:
https://cdap.atlassian.net/browse/CDAP-20180
types that need some more thinking:
https://cdap.atlassian.net/browse/CDAP-20181


.github/workflows/build.yml
```diff
@@ -71,6 +71,7 @@ jobs:
     - name: Install UI Dependencies
       run: |       
         npm install -g yarn # Install Yarn globally
+        yarn set version 3.2.0
         yarn --frozen-lockfile
```
sumengwang: I don't think yarn 2 has `--frozen-lockfile`, probably should be `--immutable`

seanfeld: yeah looks like it - https://yarnpkg.com/cli/install

PR #853, opened 2022-12-16
Recipe List component for Recipe management UIs

radhikav1:
# Recipe List component for Recipe management UIs

## Description
* Added a component RecipeList which can be used at different places as below : 
* This component can be configurable to be used at 3 places in Recipe management UIs      
** Last\recently updates recipes in home page
** ViewAll recipes 
** on click of Import recipe 
* Added new Route for Recipes which shows all the recipes
* The same component is used to show in new wrangle home page

Files to ignore for review:
* app/cdap/components/RecipeManagement/RecipeImportButton.tsx
* files under **__test__** folder.

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20052](https://cdap.atlassian.net/browse/CDAP-20052)
 [CDAP-20051](https://cdap.atlassian.net/browse/CDAP-20051)

## Test Plan
Can't write e2e tests at this point of time as recipe creation UIs are not yet ready.
Unit tests are in-progress

## Screenshots
**RecipeList component**
image

**View All Recipes with Route <_http://<instance:port>/cdap/ns/default/recipes_>**
image

**RecipeList in Wrangle home page:** Please refer Saved Recipes in below screen-shot


app/cdap/components/RecipeList/reducer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export function prevPage(dispatch, state) {
+  const { previousTokens } = state;
+  if (!previousTokens.length) {
+    return;
+  }
+  dispatch({
+    type: Actions.prevPage,
+  });
+  getSavedRecipes(dispatch, { ...state, pageToken: previousTokens[previousTokens.length - 1] });
```
sumengwang: Can you combine the two dispatches into one dispatch? maybe write a `getPrevPageRecipes` function. In there you call the api with prevPageToken and dispatch

app/cdap/components/RecipeList/reducer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export function nextPage(dispatch, state) {
+  const { nextPageToken } = state;
+  if (!nextPageToken) {
+    return;
+  }
+  dispatch({
+    type: Actions.nextPage,
+  });
+  getSavedRecipes(dispatch, { ...state, pageToken: nextPageToken });
```
sumengwang: same here

app/cdap/components/RecipeList/reducer.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  dispatch({
+    type: Actions.setSort,
+    payload: {
+      sortColumn: columnName,
+      sortOrder,
+    },
+  });
+  getSavedRecipes(dispatch, { ...state, sortColumn: columnName, sortedOrder: sortOrder });
```
sumengwang: same here

app/cdap/components/RecipeList/RecipesTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+  isShowAllColumns,
+  isShowActions,
+  enableSorting,
+  viewHandler,
+  editHandler,
+  selectHandler,
+  state,
+  dispatch,
+}: IRecipeTableProps) => {
+  const renderSortableHeaderColumn = (columnName: string) => {
```
sumengwang, line 49: I'm not sure using one `sortOrder` makes sense. I feel like you should have `sortNameOrder` and `sortUpdatedOrder`. @seanfeld wdyt

sumengwang (approved):
This is great work.

app/cdap/components/RecipeList/RecipeTableRow.tsx
```diff
@@ -0,0 +1,202 @@
...
+  dispatch,
+}: IRecipeTableRowProps) => {
+  const [showPopover, setShowPopover] = useState(false);
+  const [isDeleteModalOpen, setDeleteModalOpen] = useState(false);
+  const [isDeleteLoading, setDeleteLoading] = useState(false);
+  const [deleteErrMsg, setDeleteErrMsg] = useState(null);
+  const [extendedDeleteErrMsg, setExtendedDeleteErrMsg] = useState(null);
+
+  const getDateID = (date: Date) => {
+    return moment(date).format(DATE_FORMAT);
```
njbriggs: Please use the formatters in [DataFormatter](https://github.com/cdapio/cdap-ui/blob/develop/app/cdap/services/DataFormatter/index.ts)

app/cdap/components/RecipeList/RecipeTableRow.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+const PREFIX = 'features.WranglerNewUI.Recipe';
+const DATE_FORMAT = 'MM-DD-YYYY HH:mm A';
+
+export const RecipeTableRow = ({
+  recipe,
+  isShowAllColumns,
+  isShowActions,
+  viewHandler,
+  editHandler,
```
njbriggs: Nit: Please give callback props the `onEvent` name pattern. `onView`, `onEdit`.

app/cdap/components/RecipeList/styles.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import React from 'react';
+import styled, { css } from 'styled-components';
+
+export const RecipeTableDiv = styled(({ ...props }) => <div {...props} />)`
```
njbriggs: Can this be
```
export const RecipeTableDiv = styled.div`...`
```
?

app/cdap/components/RecipeList/styles.tsx
```diff
@@ -0,0 +1,86 @@
...
+    width: 100%;
+    box-shadow: 0px 1px 1px rgba(0, 0, 0, 0.2), 0px 2px 2px rgba(0, 0, 0, 0.14), 0px 1px 5px rgba(0, 0, 0, 0.12);
+    background-color: #FFFFFF;
+    border-radius: 4px;
+
+    .grid.grid-container {
```
njbriggs, line 26: I don't know that we want to have deeply nested class-based styling within the styled-components block? How much work would it be to individually style these components?

njbriggs (commented):
Overall, this is a really good start. There are a few things to clean up or verify, but the approach is solid.

app/cdap/components/RecipeList/reducer.ts
radhikav1: I've tried but not able to update the state accurately. I will discuss with you to get the way to do so.

app/cdap/components/RecipeList/RecipesTable.tsx
radhikav1: {sortColumn, sortedOrder } together tells column & order for sorting on the table. This way, we can extend sorting functionality for columns as needed in future.  I think using sortNameOrder and sortUpdatedOrder would be too specific. Will discuss with you to understand more on this.

app/cdap/components/RecipeList/RecipeTableRow.tsx
radhikav1: Changed to DataFormatter APi

app/cdap/components/RecipeList/RecipeTableRow.tsx
radhikav1: Done

app/cdap/components/RecipeList/styles.tsx
radhikav1: Changed

app/cdap/components/RecipeList/styles.tsx
radhikav1: As I am using common classes for table styles, making it to individual styles components requires restructuring these components and re define the classes. So would like to take it as a separate task.

app/cdap/components/RecipeList/ViewAllRecipes.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <Box ml={4} mb={2}>
+        <Typography variant="h5" key="3">
+          {T.translate(`${PREFIX}.savedRecipes`)}
+        </Typography>
+      </Box>
+
+      <Box ml={4} mr={4}>
+        <RecipeListOuter
+          isOpen={true}
+          isShowAllColumns={true}
```
mrahanjam: nit: `isOpen` makes sense to me, but for the other two I think we'd better use `showAllColumns` and `showActions`.

app/cdap/components/RecipeList/ViewAllRecipes.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Link } from 'react-router-dom';
+import Typography from '@material-ui/core/Typography';
+import { getCurrentNamespace } from 'services/NamespaceStore';
+
+const PREFIX = 'features.WranglerNewUI.Recipe';
+
+const redirectToObj = `/ns/${getCurrentNamespace()}/wrangle`;
+
+const ViewAllRecipies = () => {
+  const viewRecipeHandler = (selectedObject: any) => {
```
mrahanjam, line 33: `onView`?

app/cdap/components/RecipeList/ViewAllRecipes.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // To do : Integrate view recipe details panel
+  };
+
+  const handleEditRecipe = (selectedObject: any) => {
```
mrahanjam, line 37: `onEdit`?

app/cdap/components/RecipeList/ViewAllRecipes.tsx
```diff
@@ -0,0 +1,74 @@
...
+          {T.translate(`${PREFIX}.savedRecipes`)}
+        </Typography>
+      </Box>
+
+      <Box ml={4} mr={4}>
+        <RecipeListOuter
+          isOpen={true}
+          isShowAllColumns={true}
+          isShowActions={true}
+          viewHandler={viewRecipeHandler}
```
mrahanjam, line 63: same as above, `onView` and `onEdit`

app/cdap/components/RecipeList/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+  };
+
+  useEffect(() => {
+    getSavedRecipes(dispatch, state);
+  }, []);
+
+  return (
+    <RecipeTableDiv
+      showallcolumns={isShowAllColumns ? 1 : 0}
```
mrahanjam: why not use `true` /`false`?

app/cdap/components/RecipeList/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+            dispatch={dispatch}
+          />
+          {showPagination && <Pagination />}
+        </>
+      )}
+    </RecipeTableDiv>
+  );
+};
+
+const RecipeListOuter = ({
```
mrahanjam: What's the role of this `Outer` component? to set default values?

app/cdap/components/RecipeList/reducer.ts
```diff
@@ -0,0 +1,176 @@
...
+        previousTokens: [],
+        nextPageToken: null,
+        ready: false,
+      };
+    case Actions.prevPage:
+      const lastTokenIdx = state.previousTokens.length - 1;
+      const lastToken = state.previousTokens[lastTokenIdx];
+      return {
+        ...state,
+        previousTokens: state.previousTokens.slice(0, lastTokenIdx),
```
mrahanjam, line 68: if you wanna remove the last element, you could do `state.previousTokens.pop()`

app/cdap/components/RecipeList/reducer.ts
mrahanjam: state updates are async, if you update `previousPageToken` and do `getSavedRecipes` immediately after while relying on `previousPageToken` value, it uses the old value for that state (pre-update). In your example above, `getSavedRecipes` is not aware of the updated result of `Actions.prevPage`. There are ways to make this into a one dispatch call. But I'll keep the api call separate into its own function just like you have now. We can discuss it further if you have questions

app/cdap/components/RecipeList/RecipesTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+            )}
+            {isShowActions && <strong></strong>}
+          </div>
+        </div>
+      );
+    }
+  };
+
+  const renderTableBody = () => {
+    if (!allRecipies || (Array.isArray(allRecipies) && allRecipies.length === 0)) {
```
mrahanjam: I don't think you need `!allRecipies || (Array.isArray(allRecipies)` check. TS should complain if it's not an array as you have it defined as an array in `IRecipeTableProps`

app/cdap/components/RecipeList/RecipesTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+        sortColumn={sortColumn}
+        sortOrder={sortedOrder}
+        columnName={columnName}
+        setSort={setSorting}
+      ></SortableHeader>
+    );
+  };
+
+  const renderTableHeader = () => {
+    if (allRecipies && allRecipies.length > 0) {
```
mrahanjam: `allRecipies.length > 0` should be enough

app/cdap/components/RecipeList/RecipesTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const { sortColumn, sortedOrder } = state;
+    const setSorting = () => {
+      setSort(dispatch, state, columnName);
+    };
+    return (
+      <SortableHeader
+        sortColumn={sortColumn}
+        sortOrder={sortedOrder}
+        columnName={columnName}
+        setSort={setSorting}
```
mrahanjam: you can directly do `setSort={() => setSort(dispatch, state, columnName)}`

app/cdap/components/RecipeList/RecipeTableRow.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const handleActionsClick = (e) => {
+    e.preventDefault();
+  };
+
+  const togglePopover = () => {
+    setShowPopover(!showPopover);
+  };
+
+  const getRecipeDetailsById = (responseHandler) => {
+    MyDataPrepApi.getRecipeById({
```
mrahanjam: it'd be better if we moved these api calls inside the reducer as well. That way we know where we make all our api calls in case of potential changes. You can import the functions her from the reducer just like you do for `reset`.

app/cdap/components/RecipeList/RecipeTableRow.tsx
```diff
@@ -0,0 +1,198 @@
...
+  };
+
+  const showDeleteConfirmation = () => {
+    setShowPopover(!showPopover);
+    toggleDeleteConfirmation();
+  };
+
+  const handleDeleteRecipe = () => {
+    setDeleteLoading(true);
+    MyDataPrepApi.deleteRecipe({
```
mrahanjam: same as above

app/cdap/components/RecipeList/ViewAllRecipes.tsx
radhikav1: Changed

app/cdap/components/RecipeList/ViewAllRecipes.tsx
radhikav1: Actual Events are raised in RecipeTableRow component and I've changed there. Rest of the places are passing handlers to the event, so kept as it is.

app/cdap/components/RecipeList/index.tsx
radhikav1: Changed

app/cdap/components/RecipeList/ViewAllRecipes.tsx
radhikav1: Changed in RecipeTableRow Component

app/cdap/components/RecipeList/index.tsx
radhikav1: Initially I was using store ,so Outer component is used to wrap with Provider.  Now as it is changed to reducer, those are removed, Thought to kept it as it is for better readability about the props and it's default values.  Not sure whether it would be fine or not. Please suggest me.

app/cdap/components/RecipeList/RecipesTable.tsx
radhikav1: Removed `Array` check.   But , without `!allRecipies` am getting null pointer exceptions, so kept it for now.

app/cdap/components/RecipeList/RecipesTable.tsx
radhikav1: Getting null exceptions without having 'allRecipies' in condition, so kept it as it is.

app/cdap/components/RecipeList/reducer.ts
radhikav1: Tried, but not working as expected.   Found from react docs https://beta.reactjs.org/learn/updating-arrays-in-state#updating-arrays-without-mutation and this is reason probably.

app/cdap/components/RecipeList/RecipesTable.tsx
radhikav1: Done

app/cdap/components/RecipeList/RecipeTableRow.tsx
radhikav1: Yeah. moved all api calls to reducer

app/cdap/components/RecipeList/RecipeTableRow.tsx
radhikav1: Yeah. moved all api calls to reducer

app/cdap/components/RecipeList/reducer.ts
radhikav1: Combined dispatches now. Please latest change set.

app/cdap/components/RecipeList/reducer.ts
radhikav1: Done

app/cdap/components/RecipeList/reducer.ts
radhikav1: Done

app/cdap/components/RecipeList/reducer.ts
```diff
@@ -0,0 +1,223 @@
...
+        nextPageToken: action.payload.nextPageToken,
+      };
+    default:
+      return state;
+  }
+};
+
+export function reset(dispatch, state) {
+  getRecipes({ ...defaultInitialState, pageLimit: state.pageLimit }).subscribe((res) => {
+    dispatch({
```
mrahanjam: I see that most of these dispatches only differ in action type. you can pass the action type to the `getRecipes` function and do the dispatching there with a generic payload. It will save you some repeated lines. for example in `reset`:
```javascript
export function reset(dispatch, state) {
  getRecipes({ ...defaultInitialState, pageLimit: state.pageLimit }, Actions.reset)
}
```

and in `getRecipes`:
```javascript
async function getRecipes(state, actionType) {
  const { pageToken, sortedOrder, sortColumn, pageLimit } = state;
  const response = await MyDataPrepApi.getRecipeList({
    context: getCurrentNamespace(),
    pageToken,
    sortBy: sortColumn,
    pageSize: pageLimit,
    sortOrder: sortedOrder,
  });

dispatch({
      type: actionType,
      payload: {
        sortColumn: columnName,
        sortOrder,
        recipes: res.values,
        nextPageToken: res.nextPageToken,
      },
    });
}
```

radhikav1: Yeah, done.  Now code looks more cleaner, thanks for pointing out.

app/cdap/components/RecipeList/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+            nextDisabled={!state.nextPageToken}
+            prevDisabled={!state.previousTokens.length}
+          />
+        </div>
+      </PaginationContainer>
+    );
+  };
+
+  const viewRecipeHandler = (selectedRecipeObject: IRecipe) => {
+    if (viewHandler) {
```
mrahanjam, line 78: Will there be scenarios where these handlers are not defined/passed?

app/cdap/components/RecipeList/index.tsx
mrahanjam: I would remove it since it seems redundant and not really adding much, it's just a wrapper that only passes the prop values. Up to you though.

mrahanjam (approved):
LGTM overall, left a couple of non-blocking comments.

app/cdap/components/RecipeList/index.tsx
radhikav1: Yes, as this component is used at different places, when it is shown in Wrangler new home page, there won't be any handlers. So added the checks here.

app/cdap/components/RecipeList/index.tsx
radhikav1: Yes, you are right. seems redundant, removed now.

PR #880, opened 2023-01-17
CDAP 19706 Recipe Steps Table

adityadivami:
# TITLE
Recipe Steps Table

## Description
Summary of changes
The Table that displays the list of recipe steps applied on the Grid
Includes the following,
1. UI for the Recipe Steps Table
2. Storybook for the Recipe Steps Table

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
https://cdap.atlassian.net/browse/CDAP-19706

## Test Plan
Not Required

## Screenshots
Screenshot 2023-02-14 at 12 26 06 PM
Screenshot 2023-02-14 at 12 26 13 PM

### DIFF
--- a/app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
+++ b/app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
@@ -1,0 +1,39 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import { ComponentMeta, ComponentStory } from '@storybook/react';
+import React from 'react';
+import RecipeStepsTable from '.';
+
+export default {
+  title: 'RecipeStepsTable',
+  component: RecipeStepsTable,
+} as ComponentMeta<typeof RecipeStepsTable>;
+
+const mockRecipe = [
+  "Delete Column 'body1'",
+  "Parse Column 'body2' with delimiter 'comma' and set first row as header",
+  "Change data type 'body2' to string",
+  "Parse Column 'body2' with delimiter 'comma'",
+];
+
+const Template: ComponentStory<typeof RecipeStepsTable> = (args) => <RecipeStepsTable {...args} />;
+
+export const RecipeStepsPanel = Template.bind({});
+
+RecipeStepsPanel.args = {
+  recipeSteps: mockRecipe,
+};
--- a/app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
+++ b/app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
@@ -14,9 +14,13 @@
  * the License.
  */
 
-import { ComponentMeta, ComponentStory } from '@storybook/react';
 import React from 'react';
-import RecipeStepsTable from '.';
+
+import { ComponentMeta, ComponentStory } from '@storybook/react';
+
+import RecipeStepsTable, {
+  RecipeStepsTableContainer,
+} from 'components/WranglerV2/RecipeStepsTable';
 
 export default {
   title: 'RecipeStepsTable',
@@ -24,10 +28,10 @@ export default {
 } as ComponentMeta<typeof RecipeStepsTable>;
 
 const mockRecipe = [
-  "Delete Column 'body1'",
-  "Parse Column 'body2' with delimiter 'comma' and set first row as header",
-  "Change data type 'body2' to string",
-  "Parse Column 'body2' with delimiter 'comma'",
+  "delete-column 'body1'",
+  "parse-column 'body2' with delimiter 'comma' and set first row as header",
+  "change-data-type 'body2' to string",
+  "parse-column 'body2' with delimiter 'comma'",
 ];
 
 const Template: ComponentStory<typeof RecipeStepsTable> = (args) => <RecipeStepsTable {...args} />;
@@ -36,4 +40,5 @@ export const RecipeStepsPanel = Template.bind({});
 
 RecipeStepsPanel.args = {
   recipeSteps: mockRecipe,
+  Container: RecipeStepsTableContainer,
 };
--- a/app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
+++ b/app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
@@ -1,0 +1,150 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+    srn: '01',
+  },
+  { id: 2, step: "Delete Column 'body'", srn: '02' },
+];
+
+export function RecipeStepsPanel(args) {
+  return <RecipeStepsTable {...args} />;
+}
+
+RecipeStepsPanel.args = {
```
seanfeld, line 41: Where did you see the example of doing it like this?

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  field: string;
+  headerName: string;
+  width: number;
+  sortable: boolean;
+  renderCell?: (params: GridRowData) => JSX.Element;
+}
+
+export interface IRecipeStepsRows {
+  id: number;
+  srn: string;
```
seanfeld: whats srn

app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
tharunpola: @seanfeld 

We are following the example given in the official documentation of the storybook,
Here is the link  - https://storybook.js.org/docs/react/get-started/whats-a-story

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
tharunpola: @seanfeld 
I have updated the srn prop to serialNumber.

app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
seanfeld: It looks like you're not following the typescript examples

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { IconButton, Typography } from '@material-ui/core';
+import { GridRowData } from '@material-ui/data-grid';
+import DeleteOutlinedIcon from '@material-ui/icons/DeleteOutlined';
+import {
+  GridContainer,
+  StyledDataGrid,
+  Wrapper,
+} from 'components/WranglerV2/RecipeStepsTable/styles';
+import React from 'react';
```
seanfeld, line 17: React comes first in import order

seanfeld: Just to be clear - external libraries should always be imported before internal components and with tsx files react should be imported first.

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      </Typography>
+      {params.colDef.showDeleteButton && (
+        <IconButton onClick={(event) => onDeleteIconClick(params.row)}>
+          <DeleteOutlinedIcon />
+        </IconButton>
+      )}
+    </Wrapper>
+  );
+
+  // 1 is the index of the cell in RecipeStepsTable that needs custom render component
```
seanfeld: I don't understand this comment? Also us `column` instead of `eachColumn` for the callback arguments in `columns.forEach`. each doesn't add any value and it makes the variable longer.

app/cdap/components/WranglerV2/RecipeStepsTable/styles.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import grey from '@material-ui/core/colors/grey';
+import { DataGrid } from '@material-ui/data-grid';
+import styled from 'styled-components';
+
+export const GridContainer = styled.div`
+  height: 188px;
+`;
+
+// the !important is used where higher specificity is not working due to pre-defined inline styles on the components
```
seanfeld: This use of !important is far too liberal here - I haven't had to use important when dealing with material ui and if data-grid forces you to use important, then maybe you shouldn't be using data grid.

Why did we add data grid to this? Give me the concrete reasons because it looks like a basic grid that could've easily been built by our current libraries and our current components.

seanfeld (changes requested):

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
adityadivami: renamed `eachColumn` to `column`

removed the unwanted comment line as it is not required

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
adityadivami: updated import statements in the proper order

app/cdap/components/WranglerV2/RecipeStepsTable/RecipeStepsTable.stories.tsx
adityadivami: updated the code according to the typescript examples

app/cdap/components/WranglerV2/RecipeStepsTable/styles.ts
adityadivami: I have removed the use of !important
I have removed the data-grid library from @material-ui and created a custom table using MUI components

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const [rows, setRows] = useState([]);
+
+  useEffect(() => {
+    const updatedSteps = recipeSteps.map((recipeStep: string, index: number) => ({
+      serialNumber: String(index),
+      recipeStep,
+    }));
+
+    setRows(updatedSteps);
+  }, []);
```
GnsP: shouldn't recipeSteps be a dependency of this effect (what happens when the prop `recipeSteps` changes)?
Also, why do we need this effect and the state rows in the first place ?

Couldn't it be written simply as 

```js
const rows = recipeSteps.map((recipeStep: string, index: number) => ({
  serialNumber: String(index),
  recipeStep,
}));
```

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+          </Typography>
+          &nbsp;
+          <BodyCell />
+        </div>
+        <IconButton onClick={handleDeleteIconClick}>
+          <DeleteOutlineIcon />
+        </IconButton>
+      </RecipeStepCellWrapper>
+    );
+  };
```
GnsP: please move this function outside the component.

GnsP (changes requested):

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
adityadivami: updated the code
removed unnecessary useState and useEffect hooks from the component
assigned the map return value directly to `const rows`

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
adityadivami: updated the function

app/cdap/components/WranglerV2/RecipeStepsTable/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+  <Typography component="span" variant="body2">
+    {value}
+  </Typography>
+);
+
+export const getTableHeaderCell = (label: string) => () => (
+  <Typography component="span" variant="body1">
+    {label}
+  </Typography>
+);
```
GnsP, line 89: These can be refactored as Components, instead of functions that return components (for simplicity). 

But this current implementation is also fine.

GnsP (approved):

PR #891, opened 2023-02-06
cdap-20173 fix replication UI loading state on assessment error

GnsP:
# [cdap-20173](https://cdap.atlassian.net/browse/CDAP-20173) fix replication UI loading state on assessment error

[fix] replication UI gets blocked in case of backend error on table assessment

## Description
Fixes the following:

```
Steps to reproduce:-
1. Start creating a SQL server replication pipeline
2. Reach Select table and transformations
3. Click on a table which has column of type datetimeoffset
4. Observe that server returns an error response, but UI continues to show in progress screen indefinitely.
```

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [20173](https://cdap.atlassian.net/browse/CDAP-20173)

## Test Plan
Manual

## Screenshots
Screenshot 2023-02-08 at 12 12 18 PM

### DIFF
--- a/app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
@@ -32,6 +32,7 @@ import {
   IColumnTransformation,
   IDeltaConfig,
 } from 'components/Replicator/types';
+import { ErrorType } from 'components/Replicator/constants';
 import { useDebounce } from 'services/react/customHooks/useDebounce';
 
 import { ISelectColumnsProps, ReplicateSelect } from './types';
@@ -164,7 +165,10 @@ const SelectColumnsView = (props: ISelectColumnsProps) => {
         });
       },
       error: (err) => {
+        dispatch({ type: 'setLoading', payload: false });
         dispatch({ type: 'setError', payload: err });
+        props.toggle();
+        props.setErrorToast(ErrorType.tableAssessmentError);
       },
       complete: () => {
         dispatch({ type: 'setLoading', payload: false });
--- a/app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
+++ b/app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
@@ -462,7 +462,7 @@ class SelectTablesView extends React.PureComponent<ISelectTablesProps, ISelectTa
       return (
         <Alert
           showAlert={!!this.state.error}
-          message={T.translate(`${PREFIX}.genericErrorMessage`)}
+          message={T.translate(`${PREFIX}.genericErrorMessage`).toString()}
           onClose={this.onErrorSnackbarClose}
           type={ALERT_STATUS.Error}
           autoCloseTimeout={5000}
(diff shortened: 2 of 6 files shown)


seanfeld:
Please put something in the title so we have a little bit more context besides 'fix cdap' so we have a little bit of context quickly - like 'cdap-20173 assessment error' or something like that

app/cdap/text/text-en.yaml
```diff
@@ -2870,6 +2870,8 @@ features:
       Content:
         SelectColumns:
           primaryKeyDescription: Must include primary key
+        SelectTables:
+          genericErrorMessage: "Something went wrong!"
```
seanfeld: remove the exclamation point

app/cdap/components/Replicator/constants.tsx
```diff
@@ -23,3 +23,8 @@ export const PROGRAM_INFO = {
   programType: 'workers',
   programId: 'DeltaWorker',
 };
+
+export enum ErrorType {
+  generic = 'generic',
```
seanfeld: is this generic type being used?

app/cdap/text/text-en.yaml
GnsP: done.

app/cdap/components/Replicator/constants.tsx
GnsP: nope, it's not used as of now. Removed it.

sumengwang:
also please change the PR title as well

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
```diff
@@ -164,7 +165,10 @@ const SelectColumnsView = (props: ISelectColumnsProps) => {
         });
       },
       error: (err) => {
+        dispatch({ type: 'setLoading', payload: false });
         dispatch({ type: 'setError', payload: err });
```
sumengwang, line 169: are there cases we don't want to set loading to false when we set Error? If there's no such case we should just set loading to false in the setError dispatch and use 1 dispatch

app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
```diff
@@ -45,6 +49,8 @@ import {
   ITableInfo,
 } from 'components/Replicator/types';
 
+const PREFIX = `features.Replication.Create.Content.SelectTables`;
```
sumengwang, line 52: while you are working on this file, can you also change other hardcoded string into i18n?

GnsP:
> also please change the PR title as well

done

app/cdap/components/Replicator/Create/Content/SelectColumnsWithTransforms/index.tsx
GnsP: There are no other cases where we are using the `setError` dispatch (on this store).
But,I think, it's better to keep these as 2 dispatches, as these are 2 independent concerns. Currently there are no cases where we don't want to set loading to false when we set Error; but that does not mean there can not be any cases when we should not set loading to false when we set error.

app/cdap/components/Replicator/Create/Content/SelectTables/index.tsx
GnsP: done

sumengwang (approved):

seanfeld (approved):

PR #898, opened 2023-02-09
CDAP-20345 Correctly configure if "Resources" should be shown for namespace creation

njbriggs:
# CDAP-20345 Correctly configure if "Resources" should be shown for namespace creation

## Description
The UI component explicitly checks for `false` so if no default value is set, the Resources tab is shown. Will be cherry-picked for 6.8.1.

## PR Type
- [X] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20345](https://cdap.atlassian.net/browse/CDAP-20345)

## Test Plan
Manually verify

## Screenshots
N/A

[CDAP-20345]: https://cdap.atlassian.net/browse/CDAP-20345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/services/ThemeHelper.ts
+++ b/app/cdap/services/ThemeHelper.ts
@@ -82,7 +82,7 @@ interface IOnePoint0SpecJSON extends IThemeJSON {
     'wrangler-data-model'?: boolean;
     'wrangler-datamodel-viewer'?: boolean;
     pipelines?: boolean;
-    pipelineStudio?: boolean;
+    'pipeline-studio'?: boolean;
     analytics?: boolean;
     'rules-engine'?: boolean;
     metadata?: boolean;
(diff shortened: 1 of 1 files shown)


seanfeld (approved):

app/cdap/services/ThemeHelper.ts
```diff
@@ -435,6 +435,7 @@ function parse1Point0Spec(themeJSON: IOnePoint0SpecJSON): IThemeObj {
       showReloadSystemArtifacts: true,
       showSqlPipeline: true,
       showCDC: false,
+      onPremTetheredInstance: false,
```
rmstar, line 438: We would want to show the "Resources" tab on HDF in both tethered and non-tethered mode. 
Should this be onPremInstance instead of onPremTetheredInstance?

njbriggs: This maps to the field `on-prem-tethered-instance` in the theme file: https://github.com/cdapio/cdap-ui/blob/develop/app/cdap/services/ThemeHelper.ts#L549 I'd like to keep those 2 names in sync, and I don't think we can change the name in the theme file because that would affect instances that are already deployed.

sumengwang (approved):

PR #934, opened 2023-03-15
[CDAP-20434] Add e2e test for pushing a pipeline from list page

arjan-bal:
# Test Push  apps from Pipeline lists page.

## Description
The test performs the following steps:
1. Create a pipeline
2. Push pipeline to remote repo
3. Check if push succeeded

We DO NOT clone the repo to verify the json is present. This check will be covered by pulling the pushed pipeline back from Git in a followup PR.

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [x] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20434](https://cdap.atlassian.net/browse/CDAP-20434)

## Screenshots

[CDAP-20434]: https://cdap.atlassian.net/browse/CDAP-20434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #935, opened 2023-03-15
update cdap.json

sumengwang:
# update cdap.json

## Description
enable ff in ui


arjan-bal (commented):
LGTM, I tested the changes in a PR that requires the frontend feature flag for SCM and it works: https://github.com/cdapio/cdap-ui/pull/934

de-lan (approved):

PR #942, opened 2023-03-16
chore(CDAP-20459): use non-directional "Connected" status in UI

de-lan:
chore(CDAP-20459): use non-directional "Connected" status in UI

Because the filehash can be generated both in pull and push, we should use non-directional wording

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links

## Screenshots
Before:
image

After: 
image

### DIFF
--- a/app/cdap/text/text-en.yaml
+++ b/app/cdap/text/text-en.yaml
@@ -3189,9 +3189,9 @@ features:
       searchLabel: Search by batch pipeline name
       tab: Local pipelines
     table:
-      connected: Connected to Git
+      connected: Connected
       pipelineName: Pipeline name
-      gitStatus: Pushed to Git
+      gitStatus: Connected to Git
       pullFail: Failed to pull this pipeline from remote.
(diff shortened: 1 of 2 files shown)


seanfeld (approved):

app/cdap/text/text-en.yaml
```diff
@@ -3189,9 +3189,9 @@ features:
       searchLabel: Search by batch pipeline name
       tab: Local pipelines
     table:
-      connected: Connected to Git
+      connected: Connected
```
sumengwang, line 3192: The chip in detail page is using this text, maybe you can change it to the below one

sumengwang (approved):

app/cdap/text/text-en.yaml
de-lan: got it, updated the other place

PR #971, opened 2023-04-08
Remove snapshot

CuriousVini:
# TITLE

## Description
Summary of changes

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <groupId>io.cdap.cdap</groupId>
     <artifactId>cdap</artifactId>
-    <version>6.9.0-SNAPSHOT</version>
+    <version>6.9.0</version>
   </parent>
 
   <artifactId>cdap-ui</artifactId>


sandboxjs/sandbox_version.json
```diff
@@ -1,4 +1,4 @@
 {
-  "release": "v6.9.0-SNAPSHOT",
-  "version": "6.9.0-SNAPSHOT"
```
de-lan: I think we can leave out this doc because it's for the UI e2e tests downloading sandbox

fyi, @sumengwang @njbriggs

de-lan: We can update it once cdap 6.9.0 is released

njbriggs: Correct, we shouldn't change this until the `6.9.0` build is available. And really, if `6.9.1-SNAPSHOT` is ready, we should go right to that.

vsethi09 (approved):

PR #996, opened 2023-04-24
6.9.1cherrypick: use Object.keys in the feature flag for loop

de-lan:
691cherrypick/add e2e optional run status


de-lan:
Holding this off, might be a backend bug: https://cdap.atlassian.net/browse/CDAP-20593

Update: only cherry pick the Object.keys change

seanfeld (approved):

PR #997, opened 2023-04-25
[CDAP-20527] (fix) Fix dataproc profile missing default value

sumengwang:
# [CDAP-20527] (fix) Fix dataproc profile missing default value

## Description
I believe the issue is the system dataproc profile only returns a small fraction of properties so that it fails to render some default value
image
This pr's solution is to fill in some default values for missing properties with the `provisioner` endpoint. 

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20527](https://cdap.atlassian.net/browse/CDAP-20527)

## Test Plan
verified locally 
image
image

[CDAP-20527]: https://cdap.atlassian.net/browse/CDAP-20527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CDAP-20527]: https://cdap.atlassian.net/browse/CDAP-20527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
+++ b/app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
@@ -204,7 +204,10 @@ export default class ProfileCustomizeContent extends PureComponent {
                   }
                   return {
                     ...property,
-                    value: this.latestValues[property.name],
+                    value:
+                      this.latestValues[property.name] ||
+                      this.props.customizations[property.name] ||
+                      property['widget-attributes'].default,
                   };
                 });
               return (


seanfeld (approved):

de-lan (approved):

app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
```diff
@@ -80,13 +86,39 @@ export default class ProfileCustomizeContent extends PureComponent {
...
+        name: property.name,
+        value: property['widget-attributes'].default ? property['widget-attributes'].default : '',
+        isEditable: true,
+      };
+    });
+    return defaultProperties;
+  };
+
+  combineAndFilterPropertiesWithDefaultValue = (originalProperties, defaultProperties) => {
+    let combinedProperties = [...originalProperties];
```
njbriggs: We're using `PropertyRow` and `AbstractWidget` under the hood, so if you pass in `default` with the widget props, will that get picked up? Would save a lot of work here.

sumengwang: Terrific suggestion! I have made the change

app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
```diff
@@ -83,6 +87,17 @@ export default class ProfileCustomizeContent extends PureComponent {
     this.latestValues = {};
   }
 
+  getDefaultPropertiesFromProvisionerDetails = (provisionerConfigGroups) => {
+    const properties = provisionerConfigGroups.flatMap((configGroup) => configGroup.properties);
+    const defaultProperties = {};
+    properties.forEach((property) => {
+      defaultProperties[property.name] = property['widget-attributes'].default
+        ? property['widget-attributes'].default
+        : '';
```
njbriggs: Why not do `defaultProperties[property.name] = property['widget-attributes'].default`? Note that there are property types other than `text` with defaults (e.g. `workerDiskGB` has default `1000`).

app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
njbriggs: The new change is much cleaner, but I was actually trying to say that I think you could send `widget-attributes`, including `default`, into `property` passed to `ProfilePropertyRow`. Then, in `StateWrapper`, the default will be used if there's no value set.

sumengwang: hmm I took a look and I believe the `widget-attributes` are already sent to `ProfilePropertyRow`
image

but it's not updating to default value when value is missing.

app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
sumengwang: All the values are in String format 
image
And when empty, it's an empty string

sumengwang: removed the unnecessary condition

app/cdap/components/PipelineDetails/ProfilesListView/ProfileCustomizePopover/ProfileCustomizeContent/index.js
sumengwang: And we also need to take care of customization value from runtimeargs, I believe it makes more sense to directly passing the value that we want to show.

sumengwang: then it made me realize the fix can be much simpler...

PR #1000, opened 2023-05-02
fix(CDAP-20605): SCM UI should not be calling validation before listing pipelines

de-lan:
fix(CDAP-20605): SCM UI should not be calling validation before listing pipelines

## PR Type
- [x] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20605](https://cdap.atlassian.net/browse/CDAP-20605)

[CDAP-20605]: https://cdap.atlassian.net/browse/CDAP-20605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/cdap/components/NamespaceAdmin/SourceControlManagement/index.tsx
+++ b/app/cdap/components/NamespaceAdmin/SourceControlManagement/index.tsx
@@ -31,7 +31,7 @@ import { ISourceControlManagementConfig } from './types';
 import SourceControlManagementForm from './SourceControlManagementForm';
 import PrimaryTextButton from 'components/shared/Buttons/PrimaryTextButton';
 import { getCurrentNamespace } from 'services/NamespaceStore';
-import { validateSourceControlManagement } from '../store/ActionCreator';
+import { getSourceControlManagement } from '../store/ActionCreator';
 import Alert from 'components/shared/Alert';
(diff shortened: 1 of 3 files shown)


seanfeld (approved):

app/cdap/components/NamespaceAdmin/SourceControlManagement/index.tsx
```diff
@@ -75,7 +75,7 @@ export const SourceControlManagement = () => {
   const validateConfigAndRedirect = () => {
     setLoading(true);
     const namespace = getCurrentNamespace();
-    validateSourceControlManagement(namespace).subscribe(
+    getSourceControlManagement(namespace).subscribe(
```
sumengwang, line 78: If we are not going to validate, we can delete the whole block and just do ```window.location.href = `/ns/${namespace}/scm/sync`;```

sumengwang: oh, I see you are using the get to mimic validating. Then I think this is fine.

sumengwang (approved):

PR #1031, opened 2023-05-23
[CDAP-20664] (feat) add daily schedule to check cdap-ui e2e tests health

sumengwang:
# [CDAP-20664] (feat) add daily schedule to check cdap-ui e2e tests health

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20664](https://cdap.atlassian.net/browse/CDAP-20664)

[CDAP-20664]: https://cdap.atlassian.net/browse/CDAP-20664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CDAP-20664]: https://cdap.atlassian.net/browse/CDAP-20664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,8 @@
 name: Build with Unit and Integration Tests
 
 on:
+  schedule:
+    - cron: '0 8 * * *'
   workflow_run:
     workflows:
       - Trigger build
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -37,7 +39,7 @@ jobs:
     steps:
     # Pinned 1.0.0 version
     - uses: marocchino/action-workflow_run-status@54b6e87d6cb552fc5f36dbe9a722a6048725917a
```
itsankit-google, line 65: I think you will need to skip this step as well when event is not workflow run otherwise workflow will fail.

sumengwang: yes i have the below condition

.github/workflows/build.yml
```diff
@@ -37,7 +39,7 @@ jobs:
...
-      if: github.event_name != 'workflow_dispatch'
+      if: github.event_name == 'workflow_run'
     - uses: actions/checkout@v3
       with:
         ref: ${{ github.event.workflow_run.head_sha }}
```
itsankit-google: We also need to fix this. I don't think it will work on scheduled run.

.github/workflows/build.yml
itsankit-google: sg

.github/workflows/build.yml
sumengwang: I tried to run workflow_dispatch. And it seems it didn't complain

sumengwang: So i believe when this is empty, it will just reference the default develop branch. I guess if I want to run it against release branch, I should do the similar matrix thing in cdap-build?

itsankit-google: Scheduled builds only run on `default` branch. If you want to run it against release branch also on schedule that you might need to build something similar hack we do in `cdap-build`.

If you only need to run it against release branch using `workflow_dispatch` that can be done by adding a step which computes the branch based on github event.

itsankit-google (approved):

seanfeld (approved):

njbriggs (approved):

PR #1087, opened 2023-08-04
[CDAP-20760] let hsts use settings from cdap config

sumengwang:
# [CDAP-20760] let hsts use settings from cdap config

## Description
To address customer issues on HDF cannot load UI through https

## PR Type
- [x] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-20760](https://cdap.atlassian.net/browse/CDAP-20760)

[CDAP-20760]: https://cdap.atlassian.net/browse/CDAP-20760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CDAP-20760]: https://cdap.atlassian.net/browse/CDAP-20760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/express.js
+++ b/server/express.js
@@ -136,6 +136,11 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
 
   if (!isModeDevelopment()) {
     const proxyBaseUrl = cdapConfig['dashboard.proxy.base.url'];
+    const hstsSettings = {
+       maxAge: parseInt(cdapConfig["hsts.max.age"]),
+       includeSubDomains: cdapConfig["hsts.include.sub.domains"] === 'true',
+       preload: cdapConfig["hsts.preload"] === 'true',
+    }
     let cspWhiteListUrls = [];
     if (proxyBaseUrl) {
       cspWhiteListUrls.push(proxyBaseUrl);
@@ -181,10 +186,7 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
             reportUri: `https://csp.withgoogle.com/csp/cdap`,
           },
         },
-        hsts: {
-          includeSubDomains: true,
-          preload: true,
-        },
+        hsts: cdapConfig["hsts.enabled"] === 'true' && hstsSettings,
         // Hub icons are cross-origin but don't supply CORS headers
         // TODO credentialless will also work but isn't supported by FF and Safari
         crossOriginEmbedderPolicy: false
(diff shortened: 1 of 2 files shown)


server/express.js
```diff
@@ -182,8 +182,9 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
           },
         },
         hsts: {
-          includeSubDomains: true,
-          preload: true,
+          maxAge: cdapConfig['hsts.max.age'],
```
seanfeld: we need to add these to the local development json config right? otherwise these will return undefined without the most up to date sandbox?

njbriggs: Thinking about this some more:

1. There should be a main flag for turning HSTS on/off. With no flags set, HSTS should be turned off. 

```
helmet({
  strictTransportSecurity: false,
})
```

2. If the flag is on but no age is set, we should have a default age of 1 year.
3. `includeSubDomains` and `preload` can be set to `true` if the flag is on.

sumengwang: the documentation website default max age is 180 days. Shall I use that value?

njbriggs:
```
See [Preloading Strict Transport Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preloading_strict_transport_security) for details. When using preload, the max-age directive must be at least 31536000 (1 year), and the includeSubDomains directive must be present. Not part of the specification.
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#directives

server/express.js
```diff
@@ -181,10 +181,13 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
             reportUri: `https://csp.withgoogle.com/csp/cdap`,
           },
         },
-        hsts: {
-          includeSubDomains: true,
-          preload: true,
-        },
+        strictTransportSecurity: cdapConfig["hsts.enabled"]
```
njbriggs: The boolean value is already parsed, right?

sumengwang: fixed

server/express.js
dli357: HSTS preload list is a hard-coded list in Chrome; if we don't intend on submitting our code for preload, maybe we can leave the preload flag off?

seanfeld (approved):

njbriggs (approved):

PR #1114, opened 2023-09-22
get sandbox version from pom

sumengwang:
# get sandbox version from pom

So that we don't need to update the sandbox_version file anymore

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

### DIFF
--- a/sandboxjs/sandbox_starter.py
+++ b/sandboxjs/sandbox_starter.py
@@ -50,11 +50,25 @@ def enable_features(path):
             file.seek(0)
             json.dump(file_data, file, indent = 2)
 
+def get_release_version():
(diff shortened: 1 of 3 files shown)


sandboxjs/sandbox_starter.py
```diff
@@ -50,11 +50,25 @@ def enable_features(path):
             file.seek(0)
             json.dump(file_data, file, indent = 2)
 
+def get_release_version():
```
seanfeld, line 53: can we add a test to this?

PR #1265, opened 2024-12-05
Upgrade node version as CDAP UI builds are failing with incompatible node version error

radhikav1:
# CDAP UI builds are failing with incompatible node version error

## Description
Summary of changes

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [x] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-21088](https://cdap.atlassian.net/browse/CDAP-21088)

## Test Plan
Build & tests should be successful.

## Screenshots
n/A

[CDAP-21088]: https://cdap.atlassian.net/browse/CDAP-21088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


itsankit-google (approved):

njbriggs:
Node 18 is already in maintenance state and will be deprecated in mid-2025 - see https://nodejs.org/en/about/previous-releases. Since we're making this change, we should go to 22 for the longest support window.

GnsP:
Updating the node version in the `frontend-maven-plugin` may not be the complete fix (please check). I think this plugin is used in the build step only. We should also upgrade the node version used in the ui-image (in the deployments).

radhikav1:
> Node 18 is already in maintenance state and will be deprecated in mid-2025 - see https://nodejs.org/en/about/previous-releases. Since we're making this change, we should go to 22 for the longest support window.

Got it, Thanks Nick.  Updated to 22 version.

njbriggs (approved):

radhikav1:
> Updating the node version in the `frontend-maven-plugin` may not be the complete fix (please check). I think this plugin is used in the build step only. We should also upgrade the node version used in the ui-image (in the deployments).

Updated node version in the build.yml as well.

GnsP (approved):

PR #1280, opened 2025-01-17
Add Studio V2 store with common, console and plugin reducers

GnsP:
# Add Studio V2 store with common, console and plugin reducers

## Description
As a part of migrating Studio to react, we start by moving the state stores to typescript and redux. 

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links

## Test Plan
Existing tests should pass.

## Screenshots
NA

### DIFF
--- a/app/cdap/api/settings.ts
+++ b/app/cdap/api/settings.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright © 2025 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import DataSourceConfigurer from 'services/datasource/DataSourceConfigurer';
+import { apiCreator } from 'services/resource-helper';
+
+const dataSrc = DataSourceConfigurer.getInstance();
+const basePath = '/namespaces/:namespace/configuration';
+const userSettingsPath = `${basePath}/user`;
+
+export const SettingsApi = {
+  getUserSettings: apiCreator(dataSrc, 'GET', 'REQUEST', `${userSettingsPath}`),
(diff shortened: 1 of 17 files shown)


app/cdap/api/settings.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright © 2024 Cask Data, Inc.
```
itsankit-google: nit: 2025 in all new files

GnsP: Yeah, will update the new files. Thanks for pointing out !

app/cdap/components/StudioV2/utils/objectUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * 
+ * @param object    This object will be deep cloned and the mutation will be applied
+ *                  on the clone. 
+ * @param mutation  Function to mutate the cloned copy of the object.   
+ *                  **IMPORTANT: It MUST mutate the cloned copy in-place.**  
+ *                  The return value of the mutation function is irrelevant (unused).
+ * 
+ * @returns         The cloned copy of the object.
+ */
+export function cloneAndApplyMutation<T>(object: T, mutation: (draft: T) => any): T {
```
njbriggs: This is fine, but it could trigger re-renders on parts that haven't changed. Another approach would be to define the `mutation` (or call them `helper`) functions to return `Partial<T>`. Then you could use the spread operator to create the new state instance like normal.

GnsP: @njbriggs
Deep cloning the state may trigger re-renders. However returning a `Partial<T>` and using spread operator would not work for mutations with `delete` operator. 

(Also, the `Partial<T>` in typescript performs a shallow partial conversion.)

I was considering the idea of using Proxy to intercept mutating operations (both set and delete) and capture them in a diff object, and then apply the diff on the actual state. 
We can use the `immer` library (specifically, the `immer.produce` function) to do this; tradeoff -- it's 3Kb in size. Alternatively, we can implement a simpler version of `immer.produce` API, supporting only property mutations on plain objects and arrays.

njbriggs: We already have `immer` in yarn.lock. We also have `immutable`. Would we be better off using one of those or even a `Map` for this piece of state?

GnsP: Then I think it would be better to use immer. It has a drop-in replacement for this function. Using `immutable` would require more changes.

PR #1296, opened 2025-02-28
[CDAP-21148] check pipeline size on change and before save / deploy

GnsP:
# Check pipeline size is less than 2MB

## Description
Summary of changes

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: [CDAP-21148](https://cdap.atlassian.net/browse/CDAP-21148)

## Screenshots

**The pipeline in this screenshot has a large description**
image

[CDAP-21148]: https://cdap.atlassian.net/browse/CDAP-21148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/hydrator/controllers/create/toppanel-ctrl.js
+++ b/app/hydrator/controllers/create/toppanel-ctrl.js
@@ -48,6 +48,7 @@ class HydratorPlusPlusTopPanelCtrl {
     HydratorUpgradeService,
     MyPollingService
   ) {
+    'ngInject';
     this.consoleStore = HydratorPlusPlusConsoleStore;
     this.myPipelineExportModalService = myPipelineExportModalService;
     this.HydratorPlusPlusConfigStore = HydratorPlusPlusConfigStore;
@@ -588,6 +589,7 @@ class HydratorPlusPlusTopPanelCtrl {
     this.checkNameError();
   }
   onPublishV2(isEdit = false) {
+    if (!this.HydratorPlusPlusConfigStore.checkPipelineJsonSize()) return;
     this.HydratorPlusPlusConfigActions.publishPipeline(isEdit);
     this.checkNameError();
   }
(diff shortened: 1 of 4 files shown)


app/hydrator/controllers/create/toppanel-ctrl.js
```diff
@@ -1456,6 +1457,16 @@ class HydratorPlusPlusTopPanelCtrl {
     }
 
     let uploadedFile = files[0];
+
+    const MB = 1024 * 1024; // Bytes
+    if (files[0].size > 2 * MB) {
```
itsankit-google: can we make it configurable from CConf?

GnsP: Yeah, we can. Would require backend changes also. We would need the value inside CDAP_CONFIG.

GnsP: For now, we are going to to keep this size limit as a static value in the UI.

sidhdirenge: Let's make it configurable from cConf if possible?
We are planning to introduce chunking soon, so UI check can become a blocker during testing

itsankit-google: > Yeah, we can. Would require backend changes also. We would need the value inside CDAP_CONFIG.

I don't think we need the changes in backend for this.

GnsP: yes, we do. Here is the corresponding backend PR: https://github.com/cdapio/cdap/pull/15921

server/config/development/cdap.json
```diff
@@ -23,5 +23,6 @@
   "hsts.enabled": "false",
   "hsts.max.age": 31536000,
   "hsts.include.sub.domains": "true",
-  "hsts.preload": "true"
+  "hsts.preload": "true",
+  "app.cdap.file.max.size.bytes": "2097152"
```
itsankit-google, line 27: why do we have the default value here if it is now coming from backend?

This is an anti-pattern; default values should be configured in a single place, either in the UI or the backend, to ensure consistency and maintainability.

GnsP: This is not a default value, these are just for development.

itsankit-google (approved):

radhikav1 (approved):

PR #1356, opened 2025-08-20
Removing snapshot for CDAP 6.11.1 release

123-komal:


adrikagupta (changes requested):
You have removed snapshot from develop branch. You need to remove it in 6.11 branch.

PR #1391, opened 2026-04-10
Restrict UI theme file loader to JSON files under the theme config dir

TristanInSec:
## Summary

`server/uiThemeWrapper.js` currently loads the UI theme file by passing the caller-supplied path straight to `__non_webpack_require__`. The `POST /updateTheme` handler in `server/express.js` feeds this function a path taken from the request body, so the effective contract is "load any file reachable by `require()` and expose its contents to the UI through `window.CDAP_UI_THEME`".

This PR narrows that contract:

- **Allowlist the theme directory.** A new `resolveAllowedThemePath()` helper canonicalises the supplied path (keeping the existing `server/` and `config/` relative-path shortcuts) and refuses to return anything that does not resolve to a location inside one of the UI theme config directories.
- **Require a `.json` extension.** The resolver rejects paths with any other extension, so the loader can no longer be pointed at a `.js` module.
- **Switch from `require()` to `fs.readFileSync` + `JSON.parse`.** Even if the allowlist check were ever weakened, the loaded file is no longer executed as a Node module.

The exported `extractUITheme(cdapConfig, uiThemePath)` signature and return shape are unchanged; the two existing callers (`extractUIThemeWrapper` and the `/updateTheme` express handler) continue to work with the legitimate theme paths shipped in `server/config/themes/` (`default.json`, `light.json`).

## Testing

I exercised the new resolver in isolation against the server's `config/themes` directory and confirmed:

| Input | Expected | Actual |
|---|---|---|
| `config/themes/light.json` | accept | accepted |
| absolute path under `server/config/themes/` | accept | accepted |
| `../../../etc/passwd` | reject | rejected (not .json) |
| `/etc/cdap/conf/cdap-site.json` | reject | rejected (outside allowed root) |
| `config/themes/evil.js` | reject | rejected (extension) |
| `config/themes/../../etc/passwd.json` | reject | rejected (outside allowed root after resolve) |
| `''` | reject | rejected (missing) |

`default.json` and `light.json` continue to load identically to the previous implementation.

### DIFF
--- a/server/uiThemeWrapper.js
+++ b/server/uiThemeWrapper.js
@@ -24,6 +24,63 @@ const CDAP_DIST_PATH = path.normalize(__dirname + '/../public/cdap_dist');
 const log = log4js.getLogger('default');
 const uiThemePropertyName = 'ui.theme.file';
 
+// Directories from which UI theme files are allowed to be loaded.
+// extractUITheme accepts a path from HTTP input (POST /updateTheme), so
+// the resolved theme file must stay inside one of these roots and must
+// be a .json file. This prevents the loader from pulling in files from
+// arbitrary locations and blocks non-JSON modules (which would otherwise
+// be executed by require()).
+const UI_THEME_ALLOWED_ROOTS = [
+  path.resolve(__dirname, 'config', 'themes'),
+  path.resolve(__dirname, '..', 'server', 'config', 'themes'),
+];
+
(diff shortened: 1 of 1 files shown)


TristanInSec:
Hi @chtyim @ajainarayanan, friendly ping on this security fix and the related #1392, #1393, #1394. CLA is passing, happy to make any changes needed. Thanks!

PR #1405, opened 2026-06-01
Restrict /updateTheme to the configured theme file and harden market host verification

adilburaksen:
## 1. `/updateTheme`: arbitrary file load via `uiThemePath`

The `/updateTheme` handler reads `req.body.uiThemePath` and passes it to `uiThemeWrapper.extractUITheme()`, which for an absolute path calls `__non_webpack_require__(uiThemePath)` — loading any absolute path on the server as a Node module. The endpoint's only legitimate use is reloading the operator-configured theme.

**Fix:** reject any `uiThemePath` that does not exactly match the configured `ui.theme.file`. This preserves the reload-theme behavior while removing the arbitrary-`require` primitive.

## 2. `isVerifiedMarketHost`: allowlist bypass via `startsWith`

`isVerifiedMarketHost` validates a market URL with `url.startsWith(configuredMarketUrl)`. A prefix check is bypassable: with a host-only configured market URL, `https://market.example.com.evil.com/...` and `https://market.example.com@evil.com/...` both pass, letting the market proxy (`/market`, `/forwardMarketToCdap`) fetch an attacker-controlled host (server-side request forgery).

**Fix:** parse both the requested and configured URLs with `URL`, require an exact `origin` match, and require the requested path to be contained within the configured base path.

### DIFF
--- a/server/express.js
+++ b/server/express.js
@@ -704,6 +704,13 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
     if (!uiThemePath) {
       return res.status(500).send('UnKnown theme file. Please make sure the path is valid');
     }
+    // Only allow (re)loading the theme file configured by the operator. Without
+    // this check, extractUITheme() would __non_webpack_require__() an arbitrary
+    // absolute path taken from the request body, letting a caller load any file
+    // on the server as a Node module.
+    if (uiThemePath !== cdapConfig['ui.theme.file']) {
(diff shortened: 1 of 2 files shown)


adilburaksen:
@GnsP mind reviewing this? It restricts /updateTheme to the configured theme file (right now the path isn't constrained) and tightens the market host verification. Front-end-only change. Happy to split the two hardening bits into separate PRs if you'd prefer to take them one at a time.

PR #1425, opened 2026-07-25
Fix arbitrary code execution and predictable session tokens in /updateTheme

herdiyana256:
Two issues in server/token.mjs and server/uiThemeWrapper.js combine into an exploitable chain through the POST /updateTheme route in server/express.js.

getSecretFromCDAPConfig() in server/token.mjs fell back to path.resolve(__dirname, 'config', 'development', 'session.secret.key') -- a file path string, not that file's contents -- used directly as key material whenever cdapConfig['session.secret.key'] isn't configured. That value is fully predictable by anyone who knows the install directory, so a session token can be forged with no access to any real secret. The fallback also doesn't match any real file on disk (the actual file is session_secret.key, with an underscore), which points to this being a missing fs.readFileSync() call rather than intentional behavior.

extractUITheme() in server/uiThemeWrapper.js loaded the theme file with __non_webpack_require__, which executes .js files as Node modules. POST /updateTheme passes req.body.uiThemePath straight into extractUITheme() with no validation beyond the (forgeable) session-token check above. Together: forge a token, then point uiThemePath at any .js file the server can read, and its code runs with the server process's privileges.

token.mjs now throws instead of falling back to a predictable value when session.secret.key isn't configured, so token generation/validation fails closed. uiThemeWrapper.js reads and JSON.parse()s theme files instead of require()-ing them, since they're data, not code. express.js additionally resolves uiThemePath in /updateTheme to a filename inside config/themes/ only, matching the route's own doc comment describing it as a way to switch between the shipped theme files for testing.

Added a regression test for each half: token_test.mjs confirms a missing secret makes generateToken throw and validateToken reject everything, including a token forged with the old fallback logic; the new uiThemeWrapper.test.js confirms a plain JSON theme file still loads and a .js file passed as the theme path is no longer executed.

yarn install didn't complete in the environment I used for this (large, older dependency tree), so I couldn't run uiThemeWrapper.test.js through the project's actual Jest config. Verified its logic instead by extracting extractUITheme's unmodified post-fix code into a standalone harness (only __dirname/log4js/lodash shimmed for environment reasons) and running it directly under Node -- confirmed the JSON case still loads and the .js case no longer executes. Happy to re-verify against the real Jest run if that's easier for a reviewer with the deps already installed.

### DIFF
--- a/server/express.js
+++ b/server/express.js
@@ -709,6 +709,13 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
    * be used and we need to persist this information somewhere.
    */
   app.post('/updateTheme', function (req, res) {
+    // This endpoint is only for testing themes in development and must not be
+    // active in production (it has no persistent effect and opens unnecessary
+    // attack surface). Return 401 so callers get a clear signal rather than a
+    // silent no-op.
+    if (isModeProduction()) {
+      return res.status(401).send('This endpoint is not available in production');
+    }
     let authToken = req.headers.authorization;
     if (
       !req.headers['session-token'] ||
--- a/server/express.js
+++ b/server/express.js
@@ -717,10 +717,16 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
     ) {
       return res.status(500).send('Unable to validate session');
     }
-    const uiThemePath = req.body.uiThemePath;
-    if (!uiThemePath) {
+    const requestedTheme = req.body.uiThemePath;
+    if (!requestedTheme) {
       return res.status(500).send('UnKnown theme file. Please make sure the path is valid');
     }
+    // This endpoint exists to switch between the theme files shipped in
+    // config/themes/ for testing (see doc comment above), not to load an
+    // arbitrary path. Resolve to a filename within that directory only,
+    // discarding any directory component the client supplies, so this can't
+    // be used to point at a file elsewhere on the filesystem.
+    const uiThemePath = path.join(__dirname, 'config', 'themes', path.basename(requestedTheme));
     try {
       uiThemeConfig = uiThemeWrapper.extractUITheme(cdapConfig, uiThemePath);
     } catch (e) {
--- a/server/express.js
+++ b/server/express.js
@@ -718,15 +718,23 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
       return res.status(500).send('Unable to validate session');
     }
     const requestedTheme = req.body.uiThemePath;
-    if (!requestedTheme) {
+    if (typeof requestedTheme !== 'string' || !requestedTheme) {
       return res.status(500).send('UnKnown theme file. Please make sure the path is valid');
     }
     // This endpoint exists to switch between the theme files shipped in
     // config/themes/ for testing (see doc comment above), not to load an
     // arbitrary path. Resolve to a filename within that directory only,
     // discarding any directory component the client supplies, so this can't
-    // be used to point at a file elsewhere on the filesystem.
-    const uiThemePath = path.join(__dirname, 'config', 'themes', path.basename(requestedTheme));
+    // be used to point at a file elsewhere on the filesystem. path.basename
+    // returns '..' or '.' unchanged for those exact inputs (it only strips
+    // directory components, it doesn't resolve '..' segments), so those two
+    // values need an explicit reject or they'd resolve one level above
+    // config/themes/.
+    const baseTheme = path.basename(requestedTheme);
+    if (baseTheme === '..' || baseTheme === '.' || !baseTheme) {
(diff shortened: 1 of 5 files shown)


server/express.js
```diff
@@ -717,10 +717,16 @@ function makeApp(authAddress, cdapConfig, uiSettings) {
...
+    const requestedTheme = req.body.uiThemePath;
+    if (!requestedTheme) {
       return res.status(500).send('UnKnown theme file. Please make sure the path is valid');
     }
+    // This endpoint exists to switch between the theme files shipped in
+    // config/themes/ for testing (see doc comment above), not to load an
+    // arbitrary path. Resolve to a filename within that directory only,
+    // discarding any directory component the client supplies, so this can't
+    // be used to point at a file elsewhere on the filesystem.
+    const uiThemePath = path.join(__dirname, 'config', 'themes', path.basename(requestedTheme));
```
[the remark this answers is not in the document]
herdiyana256: Good catch, thanks. Fixed in 3250169: reject non-string uiThemePath before it reaches path.basename, and explicitly reject when the resolved basename is '..' or '.' (path.basename returns those two unchanged, it doesn't resolve them away, so they'd otherwise land one directory above config/themes/).

server/token.mjs
```diff
@@ -97,15 +96,20 @@ function decrypt(encText, key) {
 }
 
 function getSecretFromCDAPConfig(cdapConfig, logger) {
-  let secretKey =
-    cdapConfig['session.secret.key'] ||
-    path.resolve(__dirname, 'config', 'development', 'session.secret.key');
```
GnsP, line 102: Good catch !

I would suggest that we fallback to `fs.readFileSync(path.resolve(__dirname, 'config', 'development', 'session_secret.key'), 'utf8');` when the app is running with `NODE_ENV=development`.

server/express.js
GnsP, line 711: This API should not be available in production in the first place.
If the UI app is running with NODE_ENV=production, then this endpoint should return 401.

server/uiThemeWrapper.js
```diff
@@ -50,6 +50,15 @@ function extractUIFeaturesFromConfig(cdapConfig) {
   return featuresMap;
 }
 
+// Theme files are plain JSON data, not code. Read and parse them directly
+// instead of require()'ing them: __non_webpack_require__ executes .js files
+// as Node modules, which turned an attacker-controlled uiThemePath (see
+// express.js's /updateTheme route) into arbitrary code execution rather than
+// just a theme-config read.
```
GnsP: This comment can be removed after we disable /updateTheme in production.

GnsP (changes requested):

server/token.mjs
herdiyana256: Done in f77678d. When NODE_ENV=development and session.secret.key is not set, it now reads the file contents with fs.readFileSync from config/development/session_secret.key. Production still throws.

server/express.js
herdiyana256: Fixed in f77678d. Added a production guard at the top of the handler, returns 401 immediately when NODE_ENV=production.

server/uiThemeWrapper.js
herdiyana256: Removed in f77678d, after the production guard was added.

herdiyana256:
All three points addressed in f77678d, threads resolved.

herdiyana256:
@GnsP All threads addressed, and I also removed an unused `fileURLToPath` import in `server/token.mjs` that snuck into f77678d (would've failed `no-unused-vars`). Ready for re-review whenever you have a moment.

GnsP (approved):

GnsP:
Hi @herdiyana256,

Thank you so much for discovering this vulnerability and taking the time to contribute a solid fix! We deeply appreciate the effort you put into helping keep this project secure.

Also, thank you for your quick turnaround on the changes we discussed during the review. The code looks good to me, and I've just approved the PR so we can get this merged.

We would be absolutely thrilled to see more contributions from you in the future! Whether it's more security improvements, general bug fixes, or new features, your work is very welcome here.

Thanks again!

Best

herdiyana256:
Thanks again @GnsP, really appreciate the kind words and the quick reviews throughout.

Since this is approved with all checks green and no conflicts with `develop`, is there anything left on my end before it can land, or is it just waiting on a committer to merge? Happy to rebase or adjust anything if that helps.

Would love to get this one in, and I've got #1426 and #1427 up as well whenever you have bandwidth for those.

herdiyana256:
Hi @GnsP  gentle bump on this one.

It's been approved for two weeks now with all checks green and no conflicts with `develop`. Is there a particular committer I should ping to get it merged, or is there a release window it's waiting on?

Same question applies to #1426 and #1427 whenever you have bandwidth. Thanks!

